PR #2, opened 2021-09-13
update ignore to include IDE and OS specific file exclusions

acornett21:
update ignore to include IDE and OS specific file exclusions


jomkz:
Thanks for contributing!

Issue #5, opened 2021-11-01
Deploy the operator certification pipeline resources


jomkz:
The operator needs to deploy the operator certification pipeline resources inline with the instructions in step 6 from the Operator Certification CI Pipeline Instructions document.

https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#step6

acmenezes:
PR #13 for this just got out of the oven.

PR #13, opened 2021-11-02
ADD reconciler for pipeline and task resources - step 6 on operator-pipelines docs

acmenezes:
In a nutshell this PR solves the bare minimum on issue #5.

Added on new the dependencies.go file in package controllers:
1) Clones the repo operator-pipelines to the tmp directory in the container.
2) Reads and applies both pipelines and tasks found on the paths described in the instructions;
3) Removes the repo clone from the tmp directory in the container.

OBS:
On Step 2:

The apply part is done via dynamic client-go with unstructured objects "rest mapped" to known GVKs such as pipelines and tasks that belong to tekton.dev apigroup; That requires that the openshift-pipeline operator aka tekton be installed before. This task doesn't seem to require any other to complete therefore I put it right after the pipeline operator subscription one on the resources.go file.

I'm assuming those will be deployed on the same namespace as the CR here.

The image I used for testing this on my AWS OCP environment is on my own repo for now as I don't know yet what tagging strategy for testing images I should use on opdev. You can find it here: quay.io/acmenezes/operator-certification-operator:0.0.1-2.

We may need unit tests, comments and more logging on it (maybe on the next iteration). Glad to discuss what is the level of preference on each one of those that this team prefers.


controllers/dependencies.go
```diff
@@ -0,0 +1,168 @@
...
+	TASKS_MANIFESTS_PATH    = "ansible/roles/operator-pipeline/templates/openshift/tasks"
+)
+
+func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.ObjectMeta) error {
+
+	// Cloning operator-pipelines project to retrieve pipelines and tasks
+	// yaml manifests that need to be applied beforehand
+	// ref: https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#step-6---install-the-certification-pipeline-and-dependencies-into-the-cluster
+
+	repo := "https://github.com/redhat-openshift-ecosystem/operator-pipelines.git"
```
acornett21, line 39: Could this be a constant as well?

acmenezes: Fair enough. My bad. I think I passed too fast my eyes on that. I just changed it.

controllers/dependencies.go
```diff
@@ -0,0 +1,167 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	"sigs.k8s.io/controller-runtime/pkg/log"
```
samira-barouti, line 24: I suggest we take the preflight path and use logrus for logging. We can configure the logging options once and reuse it.

controllers/dependencies.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		log.Log.Info("Couldn't iterate over operator-pipelines yaml manifest files")
+		return err
+	}
+
+	// Removing cloned project
+	if err = os.RemoveAll(PIPELINE_MANIFESTS_PATH); err != nil {
```
samira-barouti, line 80: I think removing the manifests should be a deferred function after `git.PlainClone`. Otherwise, if any error encounters, the program execution will never reach here.
My other question is why we don't remove the files under TASK_MANIFEST_PATH?

controllers/dependencies.go
```diff
@@ -0,0 +1,167 @@
...
+		log.Log.Info("Couldn't remove operator-pipelines directory")
+		return err
+	}
+	return nil
+}
+
+func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace string) error {
+
+	b, err := ioutil.ReadFile(fileName)
```
samira-barouti, line 89: ioutil is going to be deprecated soon (I think). It would be better to use `os` package instead.

controllers/dependencies.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	dd, err := dynamic.NewForConfig(config)
+	if err != nil {
+		log.Log.Info("Couldn't initialize dynamic k8s client from config.")
+		return err
+	}
+
+	// Decoding yaml files to resource objects and apply to cluster
+	decoder := yamlutil.NewYAMLOrJSONDecoder(bytes.NewReader(b), 100)
```
samira-barouti, line 114: I was having some fun today to resolve the same issue, and I think using yaml.Unmarshal is cleaner. This is the snippet of my code to read a csv file:
```
	yfile, err := os.ReadFile("csv.yaml")
	if err != nil {
		log.Error(err)
	}

	var data operatorv1alpha1.ClusterServiceVersion
	err = yaml.Unmarshal(yfile, &data)
```
I checked the pipeline and task folders and we need to handle two CRDs (Task and Pipeline). I think this could be simplified. Please take a look at how we manage a similar situation in preflight (but it is the opposite! we try to dump go struct into yaml manifest). https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/1a2ef8a3829ea911356cd41a0171130d568379cd/certification/internal/policy/operator/deployable_by_olm.go#L378

Please let me know if we need to discuss further.

samira-barouti (changes requested):

controllers/dependencies.go
acmenezes: About logging as I put on the PR message I would suggest we discuss on further iterations. Here I'm following what I found on other pieces of code on this project itself. I would suggest creating an issue to standardize logging through the whole project. And then we can certainly include logrus as the standard.

controllers/dependencies.go
acmenezes: That's fair. My mistake. It's not PIPELINE_MANIFESTS_PATH it's the whole REPO_CLONE_PATH. And the defer is a good suggestion already done.

controllers/dependencies.go
acmenezes: Not sure about ioutil. But no problem os is now the one on the code. They have exactly the same signature.

controllers/dependencies.go
acmenezes: As it is it's working. Are we aiming for refactoring here? Wouldn't it be interesting having another iteration with another issue to simplify it? I would suggest that we unblock other tasks that depends on this step first and take the code refactoring and new approaches on another issue. Glad to jump in and make the changes there. What do you think?

controllers/dependencies.go
```diff
@@ -0,0 +1,170 @@
...
+		Progress: os.Stdout,
+	})
+	if err != nil {
+		log.Log.Info("Couldn't clone the repository for operator-pipelines.")
+		return err
+	}
+	defer r.RemovePipelineDependencyFiles(REPO_CLONE_PATH)
+
+	// Reading pipeline manifests and applying to cluster
+	root := REPO_CLONE_PATH + PIPELINE_MANIFESTS_PATH
```
bcrochet, line 50: These two blocks could be a loop.

controllers/dependencies.go
```diff
@@ -0,0 +1,170 @@
...
+	})
+	if err != nil {
+		log.Log.Info("Couldn't iterate over operator-pipelines yaml manifest files")
+		return err
+	}
+
+	return nil
+}
+
+func (r *OperatorPipelineReconciler) RemovePipelineDependencyFiles(filePath string) error {
```
bcrochet, line 82: This doesn't seem like it needs to be exported. 

removePipelineDependencyFiles

bcrochet (changes requested):
A couple more comments.

controllers/dependencies.go
acmenezes: done.

controllers/dependencies.go
acmenezes: Can you clarify? What do you mean?

bcrochet: paths := []string{PIPELINE_MANIFESTS_PATH, TASKS_MANIFESTS_PATH}
for _, path := range paths {
   ...
}

The blocks are identical. The only difference is the two paths.

bcrochet:
> > Two things...
> > 
> > 1. Unit tests?
> > 2. Please collapse all the commits. Unless a commit can stand on its own, it shouldn't be a separate commit. I see some of the individual commits have a bunch of commented out code. That definitely doesn't need to be in the repo as a part of the history.
> 
> @bcrochet
> 
>     1. about the unit tests I also wrote on the PR note. We didn't discuss on our first meeting how it would be done and with what libraries on this project. I suggest we take it on another issue and have it defined with everyone's participation.

We've used ginkgo/gomega on preflight. But we've also accepted plain old Go tests. Use whatever you're comfortable with. It's a bigger debt to have to add a test later, than to possibly migrate a test to a different framework.

> 
>     2. About the multiple commits what do you suggest here? Do you mean git rebase -i head n? Or do you apply another method to accomplish this? Sorry about my ignorance here.

My typical workflow when working on a PR like this is to `git commit --amend` whenever I'm fixing up a PR based on feedback. But yes, a `git rebase -i HEAD~n` and squash them all to a single commit. It's not completely forbidden to have multiple commits in a PR, but unless a commit adds a significant contribution to the git history, it doesn't need to be a part of main, i.e. we don't need to see your development iterations in the final product. :)

controllers/dependencies.go
samira-barouti: Yes refactoring.. I understand this is working, but I think it is too complex to follow and some of these lines could be removed. I'd personally like to push a code that is easy to maintain the first time. The refactoring is usually more complicated as you need to have/add unit tests to ensure the refactored code is not breaking, and we don't have unit tests yet.

BTW, have you seen this?
https://github.com/kubernetes/client-go/issues/193#issuecomment-343138889

acmenezes: Hi Samira, nope. I didn't see this one. But just realized that going the generic approach with client-go was not necessary there. So I went on the specific approach which is actually way easier. I just flipped to controller runtime client and that's much better now I think. Please check it out.

controllers/dependencies.go
acmenezes: Done.

acmenezes:
> > > Two things...
> > > 
> > > 1. Unit tests?
> > > 2. Please collapse all the commits. Unless a commit can stand on its own, it shouldn't be a separate commit. I see some of the individual commits have a bunch of commented out code. That definitely doesn't need to be in the repo as a part of the history.
> > 
> > 
> > @bcrochet
> > ```
> > 1. about the unit tests I also wrote on the PR note. We didn't discuss on our first meeting how it would be done and with what libraries on this project. I suggest we take it on another issue and have it defined with everyone's participation.
> > ```
> 
> We've used ginkgo/gomega on preflight. But we've also accepted plain old Go tests. Use whatever you're comfortable with. It's a bigger debt to have to add a test later, than to possibly migrate a test to a different framework.
> 
> > ```
> > 2. About the multiple commits what do you suggest here? Do you mean git rebase -i head n? Or do you apply another method to accomplish this? Sorry about my ignorance here.
> > ```
> 
> My typical workflow when working on a PR like this is to `git commit --amend` whenever I'm fixing up a PR based on feedback. But yes, a `git rebase -i HEAD~n` and squash them all to a single commit. It's not completely forbidden to have multiple commits in a PR, but unless a commit adds a significant contribution to the git history, it doesn't need to be a part of main, i.e. we don't need to see your development iterations in the final product. :)

Hi @bcrochet , so on the unit test I believe after we talked yesterday you won't them anymore now. About the commits agreed. I pass to that practice from now on. I would suggest to squash them on merge when this PR gets approved.

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+		root := REPO_CLONE_PATH + path
+
+		// walking through the each directory (pipelines or tasks)
+		err = filepath.Walk(root, func(filePath string, info os.FileInfo, err error) error {
+
+			// For each file NOT directories
+			if !info.IsDir() {
+				var pipeline tekton.Pipeline
+				var task tekton.Task
+				fmt.Print(path)
```
samira-barouti, line 55: this seems to be a leftover from the debugging. Could be removed.

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+
+				// apply pipeline yaml manifests
+				if path == PIPELINE_MANIFESTS_PATH {
+					if errors := r.applyManifests(filePath, meta.Namespace, &pipeline); errors != nil {
+						return errors
+					}
+
```
samira-barouti, line 62: please remove the extra lines. My OCD is kicking in ;)

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+	// Cloning operator-pipelines project to retrieve pipelines and tasks
+	// yaml manifests that need to be applied beforehand
+	// ref: https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#step-6---install-the-certification-pipeline-and-dependencies-into-the-cluster
+
+	_, err := git.PlainClone(REPO_CLONE_PATH, false, &git.CloneOptions{
+		URL: OPERATOR_PIPELINES_REPO,
+		// Progress: os.Stdout,
+	})
+	if err != nil {
+		log.Log.Info("Couldn't clone the repository for operator-pipelines: " + err.Error())
```
samira-barouti, line 37: This should be log.Log.Error

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+				} else {
+					if errors := r.applyManifests(filePath, meta.Namespace, &task); errors != nil {
+						return errors
+					}
+				}
+			}
+			return nil
+		})
+		if err != nil {
+			log.Log.Info("Couldn't iterate over operator-pipelines yaml manifest files: " + err.Error())
```
samira-barouti, line 73: same here

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+	return nil
+}
+
+func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace string, obj client.Object) error {
+
+	b, err := os.ReadFile(fileName)
+	if err != nil {
+		log.Log.Info("Couldn't read manifest file " + err.Error())
```
samira-barouti, line 92: same here

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+
+	if err = yamlutil.Unmarshal(b, &obj); err != nil {
+		return err
+	}
+
+	obj.SetNamespace(Namespace)
+	if err = r.Client.Create(context.Background(), obj); err != nil {
+		return err
```
samira-barouti, line 102: please log a message

samira-barouti (changes requested):

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace string, obj client.Object) error {
+
+	b, err := os.ReadFile(fileName)
+	if err != nil {
+		log.Log.Info("Couldn't read manifest file " + err.Error())
+		return err
+	}
+
+	if err = yamlutil.Unmarshal(b, &obj); err != nil {
+		return err
```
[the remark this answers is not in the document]
acmenezes: Done.

controllers/dependencies.go
acmenezes: Done.

controllers/dependencies.go
acmenezes: Done.

controllers/dependencies.go
acmenezes: Done.

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+			log.Log.Info("Couldn't iterate over operator-pipelines yaml manifest files: " + err.Error())
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *OperatorPipelineReconciler) removePipelineDependencyFiles(filePath string) error {
+	if err := os.RemoveAll(filePath); err != nil {
+		log.Log.Info("Couldn't remove operator-pipelines directory")
```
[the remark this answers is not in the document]
acmenezes: Done.

controllers/dependencies.go
acmenezes: Done.

controllers/dependencies.go
acmenezes: That's also done.

controllers/dependencies.go
acmenezes: Done.

acornett21 (approved):

controllers/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+	// ref: https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#step-6---install-the-certification-pipeline-and-dependencies-into-the-cluster
+
+	_, err := git.PlainClone(REPO_CLONE_PATH, false, &git.CloneOptions{
+		URL: OPERATOR_PIPELINES_REPO,
+	})
+	if err != nil {
+		log.Error(err, "Couldn't clone the repository for operator-pipelines")
+		return err
+	}
+	defer r.removePipelineDependencyFiles(REPO_CLONE_PATH)
```
samira-barouti, line 37: I think (and I might be wrong), that this method call should be place before the if statement on line 33. If cloning the repo fails in the middle, the control flow will return the error but not delete the created dir.

I overaly recommend 1) creating a tmp folder programmatically for the operator-pipelines (please check https://pkg.go.dev/os#MkdirTemp) to avoid unwanted surprises, and 2) move the defer method to line 33, after the git.PlainClone and before the `if err`

controllers/dependencies.go
```diff
@@ -0,0 +1,168 @@
...
+		gr, err := restmapper.GetAPIGroupResources(c.Discovery())
+		if err != nil {
+			log.Log.Info("Couldn't get API group resources")
+			return err
+		}
+
+		mapper := restmapper.NewDiscoveryRESTMapper(gr)
+		mapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		if err != nil {
+			log.Log.Info("Couldn't the preferred resource mapping for given kind.")
```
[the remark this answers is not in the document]
acmenezes: done.

controllers/dependencies.go
acmenezes: 2) I'm not sure as well if the defer function would get triggered there so just in case I moved it to line 33 as requested.
 
1) On the other hand temporary files used by programs in Linux are historically stored in the /tmp directory that happens to have, in regular systems, garbage collection in place. This is a single Linux image deployment, single process container, single user with a single purpose. It's pretty straightforward. I don't see the need to add another package/method and more complexity to address this.

samira-barouti: No problem.

samira-barouti (approved):
/lgtm

Issue #15, opened 2021-11-03
Centralized logging


samira-barouti:
I suggest we take the preflight path and use logrus for logging. We can configure the logging options once and reuse it.

_Originally posted by @samira-barouti in https://github.com/redhat-openshift-ecosystem/operator-certification-operator/pull/13#discussion_r741559139_

Issue #9, opened 2021-11-01
Allow usage of a Red Hat Container API access key for results submission


jomkz:
The operator should allow the user to provide a secret containing a Red Hat Container API access key for results submission as documented in the Operator Certification CI Pipeline Instructions.

https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#add-red-hat-container-api-access-key

acornett21:
Key to be used in CR should be `pyxis-api-secret` when creating the secrete, since this is the value the pipeline uses.

bcrochet:
Fixed by #23

Issue #8, opened 2021-11-01
Allow usage of GitHub API Token for results submission


jomkz:
The operator should allow the user to provide a secret containing a GitHub API Token for results submission as documented in the Operator Certification CI Pipeline Instructions.

https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#add-a-github-api-token-for-the-repo-where-the-pr-will-be-created

acornett21:
Key to be used in CR should be `github-api-token` when creating the secrete, since this is the value the pipeline uses.

Issue #4, opened 2021-11-01
kubeconfig secret needs valid kubeconfig


jomkz:
Currently the kubeconfig secret that is being created is empty. This secret should contain a valid kubconfig that the pipeline can use to provision resources.

skattoju:
@jomkz to what extent do we want to validate the secrets ?
- verify that a login is successful ?
- or.. just that there is some content in the secrets ?

PR #16, opened 2021-11-03
kubeconfig secret name check

skattoju:
This was meant to address #4 
* Added a kubeconfigSecretName field to the CR to specify the name of the secret containing the kubeconfig to be used by the pipeline
* Added some validation (some of it can be removed looking for comments)
* The same pattern will apply to the other two secrets i.e. github api and pyxis (Will do #8 and #9 based on feedback on this one)
* The logging may be inconsistent left it to be addressed in #15

### DIFF
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -19,8 +19,10 @@ package controllers
 import (
 	"context"
 	"errors"
-
+	"fmt"
+	certv1alpha1 "github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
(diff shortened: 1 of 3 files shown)


skattoju:
Per discussion in slack we should be using a service account for this.

controllers/secret.go
```diff
@@ -34,19 +38,65 @@ const (
...
+		}
+		pods, err := cs.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			log.Log.Error(err, "Unable to list pods using provided kubeconfig, is the kubeconfig secret valid ?")
+			return err
+		}
+		log.Log.Info(fmt.Sprintf("was able to list %d pods in all namespaces", len(pods.Items)))
+		log.Log.Info("existing kubeconfig secret seems to be usable...")
 		return nil // Existing Secret found, do nothing...
+	} else {
```
bcrochet, line 95: I'd rather see this block as:
```
if !IsObjectFound() {
    return err;
}

```
Then the rest can be outside of the if...

controllers/secret.go
```diff
@@ -34,19 +38,65 @@ const (
...
+		if err != nil {
+			log.Log.Error(err, "unable to get kubeconfig secret")
+			return err
+		}
+		if kubeConfigSecret.Data["kubeconfig"] == nil {
+			err = errors.New(fmt.Sprintf("the kubeconfig key in %s is empty!", secretName))
+			log.Log.Error(err, "The specifed kubeconfig secret does not contain a kubeconfig")
+			return err
+		}
+		config, err := clientcmd.RESTConfigFromKubeConfig(kubeConfigSecret.Data["kubeconfig"])
```
bcrochet, line 77: I'm confused as to why we would create a new client, when we have r.Client already? Can controller-runtime not be used?

bcrochet (changes requested):
Question about usage of client-go (which introduces a new dependency)

controllers/secret.go
skattoju: agreed

controllers/secret.go
skattoju: The initial thinking was to excercise the provided kubeconfig to ensure it works.. We cant make the existing client use this other kubeconfig right ?

But as discussed on slack this probably should be removed as we said we should not be "validating" the secret beyond checking that it exists..

bcrochet: You should still be able to use controller-runtime if you wanted to do that validation, but as you've said, yeah, as per the slack discussion, there's not a whole lot of value there.

skattoju:
> Question about usage of client-go (which introduces a new dependency)

removed 👍

controllers/secret.go
```diff
@@ -33,19 +35,47 @@ const (
 
 // reconcileKubeConfigSecret will ensure that the kubeconfig Secret is present and up to date.
 func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.ObjectMeta) error {
+	pipeline := types.NamespacedName{
+		Namespace: meta.Namespace,
+		Name:      meta.Name,
+	}
+	instance := &certv1alpha1.OperatorPipeline{}
```
samira-barouti, line 42: how about `pipeline` for the var name? `instance` is a bit too generic.

skattoju: already used pipeline for the NamespacedName.. pipeline and pipelineInstance ? or pipelineName and pipeline ?

samira-barouti: or opPipeline

samira-barouti (approved):
/lgtm

jomkz (approved):
/lgtm

PR #17, opened 2021-11-03
adding in support for git hub token

acornett21:
Fixes: #8 

Adding in `gitHubSecretName` to OperatorPipeline spec, which is optional so that the cluster admin can overwrite this value if they created the secret with a different name. Also, validate that this GH secret exists in the cluster.

### DIFF
--- a/api/v1alpha1/operatorpipeline_types.go
+++ b/api/v1alpha1/operatorpipeline_types.go
@@ -31,6 +31,10 @@ type OperatorPipelineSpec struct {
 	// OperatorPipelinesRelease is the Operator Pipelines release (version) to install.
 	OperatorPipelinesRelease string `json:"operatorPipelinesRelease,omitempty"`
 
+	// GitHubSecretName is the name of the secret containing the GitHub Token that will be used by the pipeline.
+	//+kubebuilder:validation:Optional
+	GitHubSecretName string `json:"gitHubSecretName,omitempty"`
+
 	// KubeconfigSecretName is the name of the secret containing the kubeconfig that will be used by the pipeline.
 	KubeconfigSecretName string `json:"kubeconfigSecretName,omitempty"`
 }
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	certv1alpha1 "github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	k8errors "k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -30,6 +32,7 @@ import (
 const (
 	KUBECONFIG_SECRET = "kubeconfig"
 	GITHUB_API_SECRET = "github-api-token"
+	GITHUB_TOKEN      = "GITHUB_TOKEN"
 	PYXIS_API_SECRET  = "pyxis-api-secret"
 )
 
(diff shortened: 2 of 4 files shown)


api/v1alpha1/operatorpipeline_types.go
```diff
@@ -30,6 +30,10 @@ type OperatorPipelineSpec struct {
 
 	// OperatorPipelinesRelease is the Operator Pipelines release (version) to install.
 	OperatorPipelinesRelease string `json:"operatorPipelinesRelease,omitempty"`
+
+	// GitHubToken is the token that will be used by the pipeline to submit a PR to Red Hat
+	//+kubebuilder:validation:Required
```
samira-barouti, line 35: From the pipeline doc, I feel this should be optional if the partners want to test the certification pipeline w/o opening new PR. I might be wrong.

controllers/secret.go
```diff
@@ -72,7 +94,7 @@ func (r *OperatorPipelineReconciler) reconcilePyxisAPISecret(meta metav1.ObjectM
 		Name:      PYXIS_API_SECRET,
 	}
 
-	secret := newSecret(key)
+	secret := newSecret(key, "", "")
```
samira-barouti, line 97: I am curious to know why we don't create an instance of `metav1.ObjectMeta` instead of `types.NamespacedName` to pass to the newSecret method? In this case we don't need to do conversion in newSecret method.

api/v1alpha1/operatorpipeline_types.go
skattoju: Isn’t the github token supposed to be a secret ? putting it in the cr would expose it ?

acornett21: @samira-barouti This is the confusing part, _this_ GH token as I understand it is to raise the PR to the Red Hat repo, the docs also elude to _another_ token that is optional for pinned digest. I feel anything in the CR of this operator should be required, since the pipeline requires these values.

@skattoju This is the part I am not sure about, how can this operator _create_ a secret without it being provided. Is there a way to say the value in the CR is a secret and not a plan text string? I'll look into that.

controllers/secret.go
acornett21: Since this method is used by all three secret types, the `NamespacedName.Name` needs to be different for each struct. The `meta` struct does not have this info. If we go with your approach, the `newSecret` method would then have four params, correct?

note: you commented on the pyxis method which isn't implemented, just needed `""` passed in bc of the refactor.

samira-barouti: Can't we modify newSecret to receive metav1.ObjectMeta, instead of `types.NamespacedName`?

samira-barouti: Per our call, the `types.NamespacesName` is required for `client.Get(context.TODO(), key, obj)` call in the controller-runtime.

samira-barouti (approved):

api/v1alpha1/operatorpipeline_types.go
skattoju: Don't think we should be creating the secret.. just checking if its there based on the secret name provided..

acornett21: My understanding is that this operator's intent is to create all the required assets necessary for the pipeline to run, that way a cluster admin does not have to create anything manually. All they have to do is install the operator and create the CR.

skattoju:
Mmm..
The cluster admin would install the operator
The users interacting with it would create CRs (and secrets)
Ideally we don't  want the users secrets exposed anymore than a k8s secret..
Given that k8s secrets are just a base64 decode away from being "exposed" I
guess the only thing protecting them is access to the namespace in which
the CR is created.. I guess it's not _that_ much of a stretch to have the
secret in the CR..
It would probably be overkill to implement some sort of way to retrieve the
token via oath or something like that.. (Unless it's a requirement)
But I believe the best practice is to reference secrets.. however it may
not be the best thing for our particular use case..

api/v1alpha1/operatorpipeline_types.go
acornett21: We only need to validate the secret exists so I have changed the code to do that and will resolve this comment.

controllers/secret.go
```diff
@@ -19,16 +19,21 @@ package controllers
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
 	KUBECONFIG_SECRET = "kubeconfig"
 	GITHUB_API_SECRET = "github-api-token"
+	GITHUB_TOKEN      = "GITHUB_TOKEN"
 	PYXIS_API_SECRET  = "pyxis-api-secret"
+	PYXIS_API_KEY     = "pyxis_api_key"
```
samira-barouti, line 36: are we using this var anywhere?

controllers/secret.go
```diff
@@ -50,18 +55,60 @@ func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.Objec
 
 // reconcileGitHubAPISecret will ensure that the GitHub API Secret is present and up to date.
 func (r *OperatorPipelineReconciler) reconcileGitHubAPISecret(meta metav1.ObjectMeta) error {
+
+	namespaceName := types.NamespacedName{
+		Namespace: meta.Namespace,
+		Name:      meta.Name,
+	}
+
+	op := &certv1alpha1.OperatorPipeline{}
```
samira-barouti, line 64: Nit: a more descriptive var name makes it easier to follow the code

controllers/secret.go
acornett21: Not yet, removed for now.

controllers/secret.go
acornett21: updated for clarity.

controllers/secret.go
```diff
@@ -80,18 +83,60 @@ func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.Objec
...
 	secret := newSecret(key)
-	if IsObjectFound(r.Client, key, secret) {
-		log.Info("existing github api secret found")
-		return nil // Existing Secret found, do nothing...
+	if !IsObjectFound(r.Client, key, secret) {
+		return errors.New(fmt.Sprintf("github api secret not found in namespace: %s", meta.Namespace))
+	}
+
+	gitHubSecret := &corev1.Secret{}
+	err = r.Get(context.Background(), key, gitHubSecret)
```
samira-barouti, line 120: In most examples that I've seen, context.TODO is used rather than `.Background()`. To me, context.TODO fits better.

controllers/secret.go
```diff
@@ -80,18 +83,60 @@ func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.Objec
...
+	if err != nil {
```
samira-barouti, line 121: lines 120 and 121 could be merged.

controllers/secret.go
```diff
@@ -80,18 +83,60 @@ func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.Objec
...
+		log.Error(err, fmt.Sprintf("unable to retrieve github api secret in namespace: %s ", meta.Namespace))
+		return err
 	}
 
-	return errors.New("github api secret not found")
+	// checking to make sure that the key is present in the map before continue to verify the data
+	if value, ok := gitHubSecret.Data[GITHUB_TOKEN]; ok {
+		if string(value) != "" {
```
samira-barouti, line 128: if value is nil, I assume this might cause nil dereference. Can we check for nil first?
IMO, we better check for the erroneous condition and let the else take care of the normal condition to avoid this sort of issue.

acornett21: value will never be nil, this is checking to make sure the `GITHUB_TOKEN` is in the map before proceeding. Checking nil on a map leads to false positives since the check for the key not being found and if the key is present but the value is nil yield the same condition met.

ie 
```
if gitHubSecret.Data[GITHUB_TOKEN] == nil {
   //do stuff
}
```
Leads to false positives.

controllers/secret.go
acornett21: How so? `err` needs to be stored to return.

samira-barouti: That's what I meant:
```if err = r.Get(context.Background(), key, gitHubSecret); err != nil {```

File: controllers/secret.go (hunk 7 above)
Comment (samira-barouti) [reply controllers/secret.go:128]: What I meant was to check for the erroneous condition as follow:
```
if value == nil || string(value) == "" {
          // handle errors
}
else {
        // the rest of code
}
```
Checking for exceptions and errors as the first step (in the if block, rather than else block) ensures that nil dereference is handled properly.

acornett21: For this case, value would never be `nil` but I will add that check in and handle errors first.

samira-barouti (approved):
/lgtm

jomkz (approved):

PR #18, opened 2021-11-04
switching to logrus logger

acornett21:
Fixes: #15 

Switching to use logrus with a centralized configuration.

### DIFF
--- a/controllers/image.go
+++ b/controllers/image.go
@@ -20,10 +20,10 @@ import (
 	"context"
 
 	imagev1 "github.com/openshift/api/image/v1"
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


controllers/image.go
```diff
@@ -20,10 +20,10 @@ import (
 	"context"
 
 	imagev1 "github.com/openshift/api/image/v1"
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/log"
```
skattoju, line 26: shouldn’t we still use the controllruntime logging interface but tell it which logging lib to use ?

something like .. 

func init() {
	log.SetLogger(logrus.New())
}

in order to take advantage of k8s specific structured logging ?

jomkz (approved):
/lgtm

controllers/image.go
acornett21: @skattoju The logs I saw with both k8log and logrus appeared the same. But I think operator sdk prefers the k8's one. This PR probably needs to change to just instantiate the logger once. I'll look more today, and test more on this.

jomkz (approved):
This seems good to me pending the comments from the others.

PR #19, opened 2021-11-04
uncommenting vendor director in ignore file so it doesn't get checked in

acornett21:
Updating the ignore file so the vendor director doesn't get accidentally checked in.


samira-barouti (approved):
/lgtm

Issue #12, opened 2021-11-02
Create the operator bundle


itroyano:
1) run `make bundle`
2) do the changes needed in manifests and kustomize
3) uncomment the build steps in GH Action

PR #20, opened 2021-11-04
initial operator bundle and build

itroyano:
Adding a CSV and bundle manifests generated by `make bundle`.
Adding builds for the bundle in GH Actions.

When building the bundle, the Dockerfile will run `make bundle`.
what this does, when bundle files already exist, is overrides the version of the controller image so that it is synced with the recently-built snapshot or release.

Fixes: #12 


Makefile
```diff
@@ -29,14 +29,14 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
...
+IMAGE_TAG_BASE ?= quay.io/opdev/operator-certification-operator
 
 # BUNDLE_IMG defines the image:tag used for the bundle.
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
-BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:$(VERSION)
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/opdev/operator-certification-operator:$(VERSION)
```
bcrochet, line 39: Can we use IMAGE_TAG_BASE here? $(IMAGE_TAG_BASE):$(VERSION)

bcrochet (changes requested):
Question about Makefile

Makefile
itroyano: We can.  but actually I would like to take another look at it because of $(VERSION).

VERSION is set to 0.0.1 at the top of the Makefile,  so installing the bundle currently won't work since there is no operator image tagged 0.0.1.  there's one with git commit.

Same applies to CSV.

Looking at ways to resolve this.  don't merge yet.

itroyano: /do-not-merge

itroyano:
label: do-not-merge/work-in-progress

itroyano:
older duplicate of https://github.com/redhat-openshift-ecosystem/operator-certification-operator/pull/27.

PR #27, opened 2021-11-09
initial operator bundle and build

itroyano:
Adding a CSV and bundle manifests generated by make bundle.
Adding builds for the bundle in GH Actions.

When building the bundle, the Dockerfile will run make bundle.
what this does, when bundle files already exist, is overrides the version of the controller image so that it is synced with the recently-built snapshot or release.

Fixes: #12

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -1,0 +1,43 @@
+ARG quay_expiration=never
+ARG release_tag=0.0.0
+
+FROM quay.io/operator-framework/operator-sdk:v1.14.0 as builder
+
+ARG release_tag
+# Copy the go source
+COPY . /workspace
+
+WORKDIR /workspace
+
+RUN microdnf install -y \
+      gcc \
+      git
+
+# Generate versioned manifests. when manifests exists this will set the version provided.
+RUN make bundle RELEASE_TAG=${release_tag}
+
+FROM scratch
+
+ARG quay_expiration
+
+# Core bundle labels.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=operator-certification-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.0
+LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
+LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v3
+
+# Labels for testing.
+LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
+LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
+
+# Define that tags should expire after 1 week. This should not apply to versioned releases.
+LABEL quay.expires-after=${quay_expiration}
+
+# Copy files to locations specified by labels.
+COPY --from=builder /workspace/bundle/manifests /manifests/
+COPY --from=builder /workspace/bundle/metadata /metadata/
+COPY --from=builder /workspace/bundle/tests/scorecard /tests/scorecard/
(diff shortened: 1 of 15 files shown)


itroyano:
/unfreeze :) 

@jomkz ready for review

bundle.Dockerfile
```diff
@@ -0,0 +1,28 @@
+ARG quay_expiration=never
+ARG release_tag=0.0.0
```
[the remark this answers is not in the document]
itroyano: that's the sneaky trick :)
make bundle on existing manifests will update the version dynamically according to what we pass here.
while the manager image will have the `VERSION` specified in makefile.

bundle.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG release_tag
+# Copy the go source
+COPY . /workspace
+
+WORKDIR /workspace
+
+RUN microdnf install -y \
+      gcc \
+      git \
+    && microdnf clean all
```
bcrochet, line 15: Nit: I don't think the clean is necessary for the builder. Feel free to ignore.

bundle.Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG quay_expiration=never
+ARG release_tag=0.0.0
+
+FROM quay.io/operator-framework/operator-sdk:v1.13 as builder
```
bcrochet, line 4: Operator-sdk is up to v1.14.0. Should we be using that?

bcrochet (commented):
Couple of questions.

config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,49 @@
...
+  keywords:
+  - certification
+  - pipeline
+  links:
+  - name: Operator Certification Operator
+    url: https://github.com/redhat-openshift-ecosystem/operator-certification-operator
+  maintainers:
+  - email: jomckenz@redhat.com
+    name: jomkz
+  maturity: alpha
```
madorn, line 45: may want to add `minKubeVersion: 1.21.0` - per [pipelines docs](https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#what-youll-need-before-you-start) we should require 4.8/1.21.0 or above.

itroyano: Since the doc states 4.8 is recommended and mentions tested with cli 4.7 - should we be strict and require 1.21.0,  or can we require 1.18.0 (OpenShift 4.7) ?

bundle.Dockerfile
itroyano: Agree.  removed.

bundle.Dockerfile
itroyano: Bumped up to 1.14.0

bundle.Dockerfile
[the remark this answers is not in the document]
itroyano: Btw it will also add api changes to the csv.
saving an untarring a recent local build - I found out it added your newest attributes from crd to the csv.
I haven't run `make bundle` after git syncing the code.   the build did - and it picked up the changes.

config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
jomkz: Yeah, the docs say " (recommended version 4.8 or above)" so I am not sure there is a hard requirement for 1.21.0 per se. @tonytcampbell Do you have any thoughts on this? Is it that we just haven't tested on anything below 4.8/1.21?

jomkz (approved):
/lgtm

bundle/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,70 @@
...
+kind: Configuration
+metadata:
+  name: config
+stages:
+- parallel: true
+  tests:
+  - entrypoint:
+    - scorecard-test
+    - basic-check-spec
+    image: quay.io/operator-framework/scorecard-test:v1.11.0
```
bcrochet, line 11: This is 3 revs back. Can we change these to v1.14.0?

bcrochet (changes requested):
Question about scorecard config...

bundle/tests/scorecard/config.yaml
itroyano: replaced across multiple yamls

bcrochet (approved):
/lgtm

Issue #31, opened 2021-11-11
Suppress `already-exists` exception for `operator-ci-pipeline`


acornett21:
## Bug Description
When reconcile happens for a 2nd time the below error happens and causes the operator to not reconcile further.

```
2021-11-11T10:15:14.370-0700    ERROR   controller_operatorpipeline     Couldn't iterate over operator-pipelines yaml manifest files    {"error": "pipelines.tekton.dev \"operator-ci-pipeline\" already exists"}
```

## Version and Command Invocation
sha of code under test `b587f2a87e0328c1d910c9cc3fa685c984d76590`

## Steps to Reproduce:

1) Run the application
2) wait for the reconcile to happy for a second time and see failure

## Expected Result
I expect that this operator suppress `already exists` exceptions to not output to the logs/console.

## Actual Result
This error is logged and throw back up the stack, thus stopping the reconciliation process

## Additional Context
N/A

dcurran90:
also seeing a problem that when the OperatorPipeline CR is deleted the pipelines do not get cleaned up. I don't believe the correct ownership is set up.
I'm thinking this needs to be a separate issue but wanted to bring it up here as it's sort of related

acornett21:
@dcurran90 I noticed that too and I was not sure if the operator should `own` any resources. But if it were to own resources, the `pipeline` would be one I could see as potentially owning. Good, to call out.

PR #34, opened 2021-11-11
adding additional error handling to not block additional reconciliation

acornett21:
- Fixes: #31 

- Calling removePipelineDependencyFiles() before each clone to make sure we always get latest code from the pipelines repo.
- Changing var name to `errs` to avoid package level collisions with the `errors` package.
- Adding separate apply() methods for pipeline/task so we can check if they exist and update them, to prevent errors when they already exist in the cluster. This also helps the operator apply new pipelines/tasks to the cluster when the pipelines repo gets updated. 

### DIFF
--- a/controllers/dependencies.go
+++ b/controllers/dependencies.go
@@ -2,21 +2,20 @@ package controllers
 
 import (
 	"context"
-
+	"fmt"
 	"os"
-
 	"path/filepath"
 
-	git "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5"
 	tekton "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	OPERATOR_PIPELINES_REPO = "https://github.com/redhat-openshift-ecosystem/operator-pipelines.git"
-	REPO_CLONE_PATH         = "/tmp/operator-pipelines/"
 	PIPELINE_MANIFESTS_PATH = "ansible/roles/operator-pipeline/templates/openshift/pipelines"
 	TASK_MANIFESTS_PATH     = "ansible/roles/operator-pipeline/templates/openshift/tasks"
 )
@@ -27,10 +26,15 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
 	// yaml manifests that need to be applied beforehand
 	// ref: https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#step-6---install-the-certification-pipeline-and-dependencies-into-the-cluster
 
-	_, err := git.PlainClone(REPO_CLONE_PATH, false, &git.CloneOptions{
+	// creating a tmp director so that each reconcile gets a new directory in case the defer does not execute properly
+	tmpRepoClonePath, _ := os.MkdirTemp("", "operator-pipelines-*")
+
+	_, err := git.PlainClone(tmpRepoClonePath, false, &git.CloneOptions{
 		URL: OPERATOR_PIPELINES_REPO,
 	})
-	defer r.removePipelineDependencyFiles(REPO_CLONE_PATH)
+
+	defer os.RemoveAll(tmpRepoClonePath)
+
 	if err != nil {
 		log.Error(err, "Couldn't clone the repository for operator-pipelines")
 		return err
@@ -40,28 +44,23 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
 	for _, path := range paths {
 
 		// base repository root + specific yaml manifest directory (pipelines or tasks)
-		root := REPO_CLONE_PATH + path
+		root := filepath.Join(tmpRepoClonePath, path)
 
-		// walking through the each directory (pipelines or tasks)
+		// walking through each directory (pipelines or tasks)
 		err = filepath.Walk(root, func(filePath string, info os.FileInfo, err error) error {
 
 			// For each file NOT directories
 			if !info.IsDir() {
-				var pipeline tekton.Pipeline
-				var task tekton.Task
 
 				// apply pipeline yaml manifests
 				if path == PIPELINE_MANIFESTS_PATH {
-					if errors := r.applyManifests(filePath, meta.Namespace, &pipeline); errors != nil {
-						log.Error(errors, "Couldn't apply pipeline manifest")
-						return errors
+					if errs := r.applyPipelineManifests(filePath, meta); errs != nil {
+						return errs
 					}
-
 					// or apply tasks manifests
 				} else {
-					if errors := r.applyManifests(filePath, meta.Namespace, &task); errors != nil {
-						log.Error(errors, "Couldn't apply task manifest")
-						return errors
+					if errs := r.applyTaskManifests(filePath, meta); errs != nil {
+						return errs
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


controllers/dependencies.go
```diff
@@ -97,7 +96,7 @@ func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace s
 	}
 
 	obj.SetNamespace(Namespace)
-	if err = r.Client.Create(context.Background(), obj); err != nil {
+	if err = r.Client.Create(context.Background(), obj); err != nil && errors.IsNotFound(err) {
```
samira-barouti, line 99: This has the risk that other errors (bad requests/conflict) will not be handled properly this way. IMO, it is better to ignore `IsAlreadyExist` error and let all the other errors to fail the method call. I mean:
```
if err = r.Client.Create(context.Background(), obj); err != nil && !errors.IsAlreadyExists(err) {
```
Ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsAlreadyExists

samira-barouti (changes requested):

controllers/dependencies.go
```diff
@@ -31,7 +30,7 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
 		URL: OPERATOR_PIPELINES_REPO,
 	})
 	defer r.removePipelineDependencyFiles(REPO_CLONE_PATH)
-	if err != nil {
+	if err != nil && err != git.ErrRepositoryAlreadyExists {
```
samira-barouti, line 33: Is PlainClone smart enough to check that the existing repo has diverged from the upstream repo? I want to make sure that we always have the most updated repo. If we ignore the AlreadyExist error AND the PlainClone does not check if the git is synced with upstream, there is a chance that we don't pull the latest code. 
If PlainClone does not check whether the git repo is synced, I'd prefer the operator to fail rather than continuing to run.

I am also wondering why we need to handle this. We have a defer method which cleans the git repo not matter what. Did you face any issue with the leftover from the previous operator execution?

samira-barouti (changes requested):

controllers/dependencies.go
acornett21: @samira-barouti This is the error I was looking for, thanks! Will update and test with this one as well.

acornett21: Ended up going a different route and updating the resource if it already exists in the cluster, since we clone the repo each reconciliation, and that could have more pipelines/tasks.

controllers/dependencies.go
acornett21: PlainClone fails is that `dir` is already on disk, sometimes reconcilation happens before the defer is finished. So changed this code to always remove the tmp `dir` which is not error prone since it does not matter if the `dir` exists or not, remove does not throw an error.

controllers/dependencies.go
```diff
@@ -96,11 +102,51 @@ func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace s
...
-		log.Error(err, "Couldn't create resource")
+	obj.SetNamespace(meta.Namespace)
+	err = r.Get(context.Background(), types.NamespacedName{Name: obj.Name, Namespace: obj.Namespace}, obj)
+	if err != nil && errors.IsNotFound(err) {
+
+		if err = r.Client.Create(context.Background(), obj); err != nil {
+			log.Error(err, "failed to create pipeline resource")
+			return err
+		}
+	} else {
```
samira-barouti, line 113: I believe this has to be more specific:
```
else if (err != nil && errors.IsAlreadyExist(err) {
   // update the resource
} else {
  return err
}
```

controllers/dependencies.go
```diff
@@ -96,11 +102,51 @@ func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace s
...
 
+	obj.SetNamespace(meta.Namespace)
+	err = r.Get(context.Background(), types.NamespacedName{Name: obj.Name, Namespace: obj.Namespace}, obj)
+	if err != nil && errors.IsNotFound(err) {
+
+		if err = r.Client.Create(context.Background(), obj); err != nil {
+			log.Error(err, "failed to create task resource")
+			return err
+		}
+	} else {
```
samira-barouti, line 144: same here

controllers/dependencies.go
```diff
@@ -27,10 +26,18 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
+	_ = r.removePipelineDependencyFiles(REPO_CLONE_PATH)
+
 	_, err := git.PlainClone(REPO_CLONE_PATH, false, &git.CloneOptions{
 		URL: OPERATOR_PIPELINES_REPO,
 	})
-	defer r.removePipelineDependencyFiles(REPO_CLONE_PATH)
+
+	defer func(r *OperatorPipelineReconciler, filePath string) {
+		_ = r.removePipelineDependencyFiles(filePath)
+	}(r, REPO_CLONE_PATH)
```
samira-barouti, line 39: I would like to have others's opinion on this too: I believe we should create a temporary dir and clone the repo into that dir. Using a non-temp dir might cause surprises if the defer method fails or does not execute for any reason.

controllers/dependencies.go
```diff
@@ -83,11 +89,11 @@ func (r *OperatorPipelineReconciler) removePipelineDependencyFiles(filePath stri
 	return nil
 }
 
-func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace string, obj client.Object) error {
+func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, meta metav1.ObjectMeta, obj *tekton.Pipeline) error {
 
 	b, err := os.ReadFile(fileName)
 	if err != nil {
-		log.Error(err, "Couldn't read manifest file")
+		log.Error(err, "Couldn't read manifest file for ")
```
samira-barouti, line 96: It would be great to print the `fileName`

samira-barouti (changes requested):

controllers/dependencies.go
acornett21: The `else` is for the `Get` not for the `Create` so if the resource exist in the cluster `obj` is populated and error is `nil`. You would be correct if it was on the `Create` but I felt that code was harder to read and overkill.

controllers/dependencies.go
acornett21: done in both methods.

controllers/dependencies.go
bcrochet: I'd agree with this. I would also say that having this be a member of the struct is probably overkill? This seems overly complicated. If it's a tmpdir, you could just 'defer os.RemoveAll(tmpdir)' and be done with it.

bcrochet (changes requested):

controllers/dependencies.go
```diff
@@ -2,16 +2,16 @@ package controllers
 
 import (
 	"context"
-
+	"fmt"
 	"os"
-
 	"path/filepath"
 
-	git "github.com/go-git/go-git/v5"
```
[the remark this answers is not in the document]
acornett21: Nice catch, seems refactor import dropped that.

controllers/dependencies.go
acornett21: Ended up making a `tmp` directory to not have to worry about directories anymore

controllers/dependencies.go
acornett21: @samira-barouti @bcrochet Please take a look at latest changes. I think this covers your concerns.

controllers/dependencies.go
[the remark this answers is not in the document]
acornett21: Added back alias.

controllers/dependencies.go
```diff
@@ -72,34 +75,67 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
+		log.Error(err, "Couldn't unmarshall yaml file")
 		return err
 	}
+
+	obj.SetNamespace(meta.Namespace)
+	err = r.Get(context.Background(), types.NamespacedName{Name: obj.Name, Namespace: obj.Namespace}, obj)
+	if err != nil && errors.IsNotFound(err) {
+
+		if err = r.Client.Create(context.Background(), obj); err != nil {
+			log.Error(err, "failed to create pipeline resource")
```
samira-barouti, line 100: would be great to have the filename in all error messages in this function

controllers/dependencies.go
```diff
@@ -52,16 +56,15 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
+						return errs
 					}
-
 					// or apply tasks manifests
 				} else {
-					if errors := r.applyManifests(filePath, meta.Namespace, &task); errors != nil {
-						log.Error(errors, "Couldn't apply task manifest")
-						return errors
+					if errs := r.applyTaskManifests(filePath, meta, &task); errs != nil {
+						log.Error(errs, "Couldn't apply task manifest")
```
samira-barouti, line 66: this could be removed. applyManifests prints enough logging messages.

controllers/dependencies.go
```diff
@@ -72,34 +75,67 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
+	if err != nil && errors.IsNotFound(err) {
+
+		if err = r.Client.Create(context.Background(), obj); err != nil {
+			log.Error(err, "failed to create pipeline resource")
+			return err
+		}
+	} else {
+		if err = r.Client.Update(context.Background(), obj); err != nil {
+			log.Error(err, "failed to update pipeline resource")
+			return err
```
samira-barouti, line 106: I feel like we don't handle the case where error is anything but IsNotFound. It will jump to the else block to update the resource, which should be avoided IMO. I think this should be the code structure:
```
err = r.Get(context.Background(), types.NamespacedName{Name: obj.Name, Namespace: obj.Namespace}, obj)

if (obj != nil) {
   // update the object and handle error
}
if (err != nil){
   if errors.IsNotFound(err) {
     // create resource
   } else { //this is error
        return err
  }
}
return nil
```

controllers/dependencies.go
```diff
@@ -52,16 +56,15 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
-					if errors := r.applyManifests(filePath, meta.Namespace, &pipeline); errors != nil {
-						log.Error(errors, "Couldn't apply pipeline manifest")
-						return errors
+					if errs := r.applyPipelineManifests(filePath, meta, &pipeline); errs != nil {
+						log.Error(errs, "Couldn't apply pipeline manifest")
+						return errs
 					}
-
 					// or apply tasks manifests
 				} else {
```
bcrochet, line 64: This else doesn't seem necessary. I think you can safely get rid of it.

controllers/dependencies.go
bcrochet: You could even get rid of the else altogether. 

if err != nil {
  if !errors.IsNotFound(err) {
    return err
  }
  createResource()
}
return nil

acornett21: I can't check for `obj !=nil` since it is already pre-populated when marshaled earlier. I'll check to see if `obj` has a `UUID` if it does it means it came from the cluster.

controllers/dependencies.go
acornett21: That is necessary, we have to create pipelines or tasks depending on what the directory is.

controllers/dependencies.go
acornett21: Done.

controllers/dependencies.go
acornett21: Done.

controllers/dependencies.go
acornett21: Took both of your suggests, please look at latest code.

samira-barouti (approved):
/lgtm pending the comments from Brad

controllers/dependencies.go
```diff
@@ -72,34 +73,79 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
 		}
 	}
+
 	return nil
 }
 
-func (r *OperatorPipelineReconciler) removePipelineDependencyFiles(filePath string) error {
-	if err := os.RemoveAll(filePath); err != nil {
-		log.Error(err, "Couldn't remove operator-pipelines directory")
+func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, meta metav1.ObjectMeta, pipeline *tekton.Pipeline) error {
```
samira-barouti, line 80: A quick question on the method signature: can we remove the third arg from both method signature and declare the var in the method body?

controllers/dependencies.go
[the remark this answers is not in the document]
bcrochet: Looks like it happened again. Bad squash?

controllers/dependencies.go
bcrochet: Oh. I see it now. It returns on error only. Got it.

controllers/dependencies.go
bcrochet: Yeah. That looks like a refactoring tool gone wrong. They're unused outside of the methods, so there is no need to pass them in, nor to return them.

acornett21: This method was 3 params to begin with but just took in OBJ, so the concrete type was defined outside the method. I agree that it makes more sense to drop the 3rd param and instantiate in the method.

controllers/dependencies.go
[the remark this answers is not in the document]
bcrochet: It seems to work. Must be an alias for it somewhere else?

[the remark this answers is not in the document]
acornett21: GoLand optimizes imports, the alias for this is still git, so it's technically unused and gets removed. But there is another package that is in `go.pkg` that also has `git` as an alias so we probably want to keep the alias to avoid collisions. To avoid both issues (collision and the alias getting dropped), I called this `gitv5` the `go.pkg` one is v4. LMK your thoughts.

samira-barouti (approved):
/lgtm

controllers/dependencies.go
```diff
@@ -72,34 +71,83 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
...
 	b, err := os.ReadFile(fileName)
 	if err != nil {
-		log.Error(err, "Couldn't read manifest file")
+		log.Error(err, fmt.Sprintf("Couldn't read manifest file for: %s", fileName))
 		return err
 	}
 
-	if err = yamlutil.Unmarshal(b, &obj); err != nil {
-		log.Error(err, "Couldn't unmarshall yaml file")
+	task := &tekton.Task{}
```
bcrochet, line 125: Nit: If you want a pointer, just use task := new(tekton.Task). Feel free to ignore. This gets the job done.

acornett21: I didn't know this was an option for structs, I changed to the more idiomatic approach. Thanks for pointing this out!

bcrochet (approved):
/lgtm

PR #35, opened 2021-11-12
refactor repeated code in secrets.go

### LINKED ISSUE #24: Refactor repeated code
Refactor repeated code resulting from #16, #17 and #23

skattoju:
fixes #24

### DIFF
--- a/config/samples/certification_v1alpha1_operatorpipeline.yaml
+++ b/config/samples/certification_v1alpha1_operatorpipeline.yaml
@@ -5,4 +5,6 @@ metadata:
 spec:
   openShiftPipelineVersion: v1.5.0
   operatorPipelinesRelease: main
+  KubeconfigSecretName: "kubeconfig"
   gitHubSecretName: "github-api-token"
+  pyxisSecretName: "pyxis-api-secret"
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -18,193 +18,114 @@ package controllers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	certv1alpha1 "github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
-	KUBECONFIG_SECRET = "kubeconfig"
-	GITHUB_API_SECRET = "github-api-token"
-	GITHUB_TOKEN      = "GITHUB_TOKEN"
-	PYXIS_API_SECRET  = "pyxis-api-secret"
-	PYXIS_API_KEY     = "pyxis_api_key"
+	defaultKubeconfigSecretName    = "kubeconfig"
+	defaultKubeconfigSecretKeyName = "kubeconfig"
+	defaultGithubApiSecretName     = "github-api-token"
+	defaultGithubApiSecretKeyName  = "GITHUB_TOKEN"
+	defaultPyxisApiSecretName      = "pyxis-api-secret"
+	defaultPyxisApiSecretKeyName   = "pyxis_api_key"
 )
 
-// reconcileKubeConfigSecret will ensure that the kubeconfig Secret is present and up to date.
-func (r *OperatorPipelineReconciler) reconcileKubeConfigSecret(meta metav1.ObjectMeta) error {
-	pipelineName := types.NamespacedName{
-		Namespace: meta.Namespace,
-		Name:      meta.Name,
-	}
-	pipeline := &certv1alpha1.OperatorPipeline{}
-	err := r.Client.Get(context.TODO(), pipelineName, pipeline)
+// ensureKubeConfigSecret will ensure that the kubeconfig Secret is present and up to date.
+func (r *OperatorPipelineReconciler) ensureKubeConfigSecret(meta metav1.ObjectMeta) error {
+	operatorPipeline, err := r.getPipeline(meta)
 	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			log.Info("pipeline resource not found. Ignoring since object must be deleted")
-			return nil
+		log.Error(err, "unable to resolve kubeconfig secret for %s/%s", meta.Namespace, meta.Name)
+		return err
+	}
+	secretName := defaultKubeconfigSecretName
+	if operatorPipeline.Spec.KubeconfigSecretName != "" {
+		secretName = operatorPipeline.Spec.KubeconfigSecretName
+	}
+	return r.ensureSecret(secretName, defaultKubeconfigSecretKeyName, meta)
+}
+
+// ensureGitHubAPISecret will ensure that the GitHub API Secret is present and up to date.
+func (r *OperatorPipelineReconciler) ensureGitHubAPISecret(meta metav1.ObjectMeta) error {
+	operatorPipeline, err := r.getPipeline(meta)
+	if err != nil {
+		log.Error(err, "unable to resolve github secret for %s/%s", meta.Namespace, meta.Name)
+		return err
+	}
+	secretName := defaultGithubApiSecretName
+	if operatorPipeline.Spec.GitHubSecretName != "" {
+		secretName = operatorPipeline.Spec.GitHubSecretName
+	}
+	return r.ensureSecret(secretName, defaultGithubApiSecretKeyName, meta)
+}
+
+// ensurePyxisAPISecret will ensure that the Pyxis API Secret is present and up to date.
+func (r *OperatorPipelineReconciler) ensurePyxisAPISecret(meta metav1.ObjectMeta) error {
+	operatorPipeline, err := r.getPipeline(meta)
+	if err != nil {
+		log.Error(err, "unable to resolve pyxis secret for %s in %s", meta.Name, meta.Namespace)
+		return err
+	}
+	secretName := defaultPyxisApiSecretName
+	if operatorPipeline.Spec.PyxisSecretName != "" {
+		secretName = operatorPipeline.Spec.PyxisSecretName
+	}
+	return r.ensureSecret(secretName, defaultPyxisApiSecretKeyName, meta)
+}
+
+// ensureSecret will ensure that the a secret with the appropriate name and key name are present
+func (r *OperatorPipelineReconciler) ensureSecret(secretName string, secretKeyName string, meta metav1.ObjectMeta) error {
+	namespacedSecretName := newNamespacedName(secretName, meta.Namespace)
+	secret := &corev1.Secret{}
+	if !IsObjectFound(r.Client, namespacedSecretName, secret) {
+		log.Error(ErrSecretNotFound, fmt.Sprintf("could not find existing secret %s/%s", meta.Namespace, secretName))
+		return ErrSecretNotFound
+	}
+	log.Info(fmt.Sprintf("found existing secret %s/%s", meta.Namespace, secretName))
+	err := r.Client.Get(context.TODO(), namespacedSecretName, secret)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("unable to get secret %s/%s", meta.Namespace, secretName))
+		return err
+	}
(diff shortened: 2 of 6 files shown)


config/samples/certification_v1alpha1_operatorpipeline.yaml
```diff
@@ -5,4 +5,6 @@ metadata:
 spec:
   openShiftPipelineVersion: v1.5.0
   operatorPipelinesRelease: main
+  KubeconfigSecretName: "kubeconfig"
   gitHubSecretName: "github-api-token"
+  PyxisApiSecretName: "pyxis-api-secret"
```
jomkz, line 10: Should this be `PyxisSecretName`?

jomkz (changes requested):

jomkz:
Sorry, just saw that this is a DRAFT. :)

config/samples/certification_v1alpha1_operatorpipeline.yaml
skattoju: yes it should 👍

config/samples/certification_v1alpha1_operatorpipeline.yaml
```diff
@@ -5,4 +5,6 @@ metadata:
 spec:
   openShiftPipelineVersion: v1.5.0
   operatorPipelinesRelease: main
+  KubeconfigSecretName: "kubeconfig"
   gitHubSecretName: "github-api-token"
+  PyxisSecretName: "pyxis-api-secret"
```
acornett21, line 10: These should be camel case to match `CSV`.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
-	GITHUB_API_SECRET = "github-api-token"
-	GITHUB_TOKEN      = "GITHUB_TOKEN"
-	PYXIS_API_SECRET  = "pyxis-api-secret"
-	PYXIS_API_KEY     = "pyxis_api_key"
+	KUBECONFIG_SECRET_NAME     = "kubeconfig"
+	KUBECONFIG_SECRET_KEY_NAME = "kubeconfig"
+	GITHUB_API_SECRET_NAME     = "github-api-token"
+	GITHUB_API_KEY_NAME        = "GITHUB_TOKEN"
+	PYXIS_API_SECRET_NAME      = "pyxis-api-secret"
+	PYXIS_API_KEY_NAME         = "pyxis_api_key"
```
acornett21, line 37: Can we either remove `secret` from the kube `key` or add `secret` to all the other `keys` so we are consistant?

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
-	pipelineInstance := &certv1alpha1.OperatorPipeline{}
-	err := r.Client.Get(context.TODO(), pipeline, pipelineInstance)
+func (r *OperatorPipelineReconciler) getPipeline(meta metav1.ObjectMeta) (*certv1alpha1.OperatorPipeline, error) {
+	pipelineName := newNamespacedName(meta.Name, meta.Namespace)
+	pipeline := &certv1alpha1.OperatorPipeline{}
+	err := r.Client.Get(context.TODO(), pipelineName, pipeline)
 	if err != nil {
-		if k8serrors.IsNotFound(err) {
+		if k8errors.IsNotFound(err) {
 			log.Info("pipeline resource not found. Ignoring since object must be deleted")
```
samira-barouti, line 116: Not sure what `Ignoring since object must be deleted` means? I've seen it in multiple places in the codebase and was wondering if this applies in the case of secrets as we don't do any CRUD operation.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
 			return nil, nil
 		}
-		log.Error(err, "unable to retrieve pipeline resource in namespace "+meta.Namespace)
+		log.Error(err, "unable to retrieve the pipeline resource %s in the namespace %s", meta.Name, meta.Namespace)
```
samira-barouti, line 119: You could probably use `namespace/name` format.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
-	err := r.Client.Get(context.TODO(), pipelineName, pipeline)
+func (r *OperatorPipelineReconciler) ensureKubeConfigSecret(meta metav1.ObjectMeta) error {
+	operatorPipeline, err := r.getPipeline(meta)
 	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			log.Info("pipeline resource not found. Ignoring since object must be deleted")
-			return nil
-		}
-		log.Error(err, "unable to retrieve pipeline resource in namespace "+meta.Namespace)
+		log.Error(err, "unable to resolve kubeconfig secret for %s in %s", meta.Name, meta.Namespace)
```
samira-barouti, line 44: You could probably use namespace/name format.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
-	KUBECONFIG_SECRET = "kubeconfig"
-	GITHUB_API_SECRET = "github-api-token"
-	GITHUB_TOKEN      = "GITHUB_TOKEN"
-	PYXIS_API_SECRET  = "pyxis-api-secret"
-	PYXIS_API_KEY     = "pyxis_api_key"
+	KUBECONFIG_SECRET_NAME     = "kubeconfig"
```
samira-barouti, line 32: Do we need to make all these values publicly available? I believe we should keep them private and follow go convention to start these vars with small letters.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
 		err := errors.New(fmt.Sprintf("an existing secret named %s was not found in namespace %s", secretName, meta.Namespace))
 		log.Error(err, fmt.Sprintf("unable to reconcile %s in namespace %s", secretName, meta.Namespace))
 		return err
 	}
-	log.Info(fmt.Sprintf("existing %s secret found in namespace %s", secretName, meta.Namespace))
-	pyxisApiSecret := &corev1.Secret{}
-	if r.Client.Get(context.TODO(), key, pyxisApiSecret) != nil {
-		log.Error(err, fmt.Sprintf("unable to get the %s secret", secretName))
+	log.Info(fmt.Sprintf("an existing secret named %s was found in namespace %s", secretName, meta.Namespace))
+	secret = &corev1.Secret{}
```
acornett21, line 91: Has anything changed from line 84? Meaning do we need to do this reassignment?

config/samples/certification_v1alpha1_operatorpipeline.yaml
skattoju: woops fixed.

controllers/secret.go
skattoju: yes can do.

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
+	err := r.Client.Get(context.TODO(), namespacedSecretName, secret)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("unable to get the %s secret in the %s namespace", secretName, meta.Namespace))
 		return err
 	}
-	if pyxisApiSecret.Data[PYXIS_API_KEY] == nil || len(pyxisApiSecret.Data[PYXIS_API_KEY]) == 0 {
-		err = errors.New(fmt.Sprintf("the pyxis api key in %s is empty!", secretName))
-		log.Error(err, fmt.Sprintf("The %s secret does not contain the expected key", secretName))
+	log.Info(fmt.Sprintf("successfully fetched secret named %s in namespace %s", secretName, meta.Namespace))
+	err = errors.New(fmt.Sprintf("the %s secret does not contain a valid key %s", secretName, secretKeyName))
```
acornett21, line 98: I prefer keeping the granularity that the original errors in the `git hub` code so we can help trouble shoot easier. If others feel that a single error for both cases is also okay, I'm okay with that as well.

controllers/secret.go
acornett21: Good call, we only need them in the `controllers` package right now.

controllers/secret.go
skattoju: Think this would apply in the case when a CR (in our case the pipeline) no longer exists and there is nothing to reconcile..

controllers/secret.go
skattoju: nope you are right its redundant

controllers/secret.go
skattoju: yes you are right seems like camel case is the go-style will change from c-style

controllers/secret.go
skattoju: yeah would be nicer

controllers/secret.go
skattoju: think they are equivalent.. both would lead to checking the value of the secret and discovering that its either not present or has a bad value..

controllers/secret.go
```diff
@@ -24,187 +24,107 @@ import (
...
-	PYXIS_API_KEY     = "pyxis_api_key"
+	defaultKubeconfigSecretName    = "kubeconfig"
+	defaultKubeconfigSecretKeyName = "kubeconfig"
+	defaultGithubApiSecretName     = "github-api-token"
+	defaultGithubApiSecretKeyName  = "GITHUB_TOKEN"
+	defaultPyxisApiSecretName      = "pyxis-api-secret"
+	defaultPyxisApiSecretKeyName   = "pyxis_api_key"
 )
 
 // reconcileKubeConfigSecret will ensure that the kubeconfig Secret is present and up to date.
```
samira-barouti: The comment need to be updated :)

controllers/secret.go
```diff
@@ -24,187 +24,108 @@ import (
...
-	if !IsObjectFound(r.Client, key, secret) {
+	return r.ensureSecret(secretName, PYXIS_API_KEY_NAME, meta)
+}
+
+func (r *OperatorPipelineReconciler) ensureSecret(secretName string, secretKeyName string, meta metav1.ObjectMeta) error {
+	namespacedSecretName := newNamespacedName(secretName, meta.Namespace)
+	secret := &corev1.Secret{}
+	if !IsObjectFound(r.Client, namespacedSecretName, secret) {
 		err := errors.New(fmt.Sprintf("an existing secret named %s was not found in namespace %s", secretName, meta.Namespace))
 		log.Error(err, fmt.Sprintf("unable to reconcile %s in namespace %s", secretName, meta.Namespace))
```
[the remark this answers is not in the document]
skattoju: replaced with ensure

controllers/secret.go
skattoju: regarding the errors file.. seems like this would be the only place we would use those errors no ? still better to separate ?

samira-barouti: It seems like we use this error twice, and seeing an error declaration in the middle of the code is a bit confusing. You could keep the error in errors.go generic, and use the error wrapping to add more context (e.g. the secret name, etc)

controllers/secret.go
```diff
@@ -24,187 +24,109 @@ import (
...
-	secret := newSecret(key)
-	if !IsObjectFound(r.Client, key, secret) {
-		return errors.New(fmt.Sprintf("github api secret not found in namespace: %s", meta.Namespace))
+// ensureSecret will ensure that the a secret with the appropriate name and key name are present
+func (r *OperatorPipelineReconciler) ensureSecret(secretName string, secretKeyName string, meta metav1.ObjectMeta) error {
+	namespacedSecretName := newNamespacedName(secretName, meta.Namespace)
+	secret := &corev1.Secret{}
+	if !IsObjectFound(r.Client, namespacedSecretName, secret) {
+		err := errors.New(fmt.Sprintf("could not find existing secret %s/%s", meta.Namespace, secretName))
+		log.Error(err, fmt.Sprintf("unable to ensure secret %s/%s", meta.Namespace, secretName))
```
samira-barouti, line 88: Same here, error wrapping could be used by defining a generic error (e.g. `unable to ensure secret`) in a file and add further context with `fmt.Sprintf("could not find existing secret %s/%s", meta.Namespace, secretName)`.

controllers/secret.go
dcurran90: the secret errors would likely only be used here but it just gives us a central location to manage all of the errors. i think that's the purpose

skattoju: yeah defining new errors in the middle is indeed jarring will start the errors file can add stuff in as we go 👍

samira-barouti (approved):
/lgtm

PR #36, opened 2021-11-12
adding checks for pinning and private registry secrets

### LINKED ISSUE #6: Add support for digest pinning
The operator should support digest pinning as documented in the Operator Certification CI Pipeline Instructions.

https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#digest-pinning-config

### LINKED ISSUE #7: Add support for private container registries
The operator should support private container registries as documented in the Operator Certification CI Pipeline Instructions.

https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md#required-step-will-be-optional-soon---if-using-a-private-container-registry

dcurran90:
Two new functions ensureDockerRegistrySecret and ensureGithubSSHSecret added to check for secrets related to private registry and digest pinning respectively.
resource.go simply logs if the secrets are not present because they are optional

Fixes #6 
Fixes #7

### DIFF
--- a/controllers/resource.go
+++ b/controllers/resource.go
@@ -68,6 +68,14 @@ func (r *OperatorPipelineReconciler) reconcileResources(ctx context.Context, pip
 		return err
 	}
 
+	if err := r.ensureDockerRegistrySecret(ctx, pipeline.ObjectMeta); err != nil {
+		log.Info("Docker Registry Secret not present or correct. Create it to use a private repository.")
+	}
+
+	if err := r.ensureGithubSSHSecret(ctx, pipeline.ObjectMeta); err != nil {
+		log.Info("Github SSH Secret not present or correct. Create it to enable digest pinning.")
+	}
+
 	if err := r.reconcileCertifiedImageStream(ctx, pipeline.ObjectMeta); err != nil {
 		return err
 	}
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -28,14 +28,18 @@ import (
 )
 
 const (
-	defaultKubeconfigSecretName    = "kubeconfig"
-	defaultKubeconfigSecretKeyName = "kubeconfig"
-	kubeConfigSecretAvailable      = "KubeConfigSecretAvailable"
-	defaultGithubApiSecretName     = "github-api-token"
-	defaultGithubApiSecretKeyName  = "GITHUB_TOKEN"
-	gitHubSecretAvailable          = "GithubSecretAvailable"
-	defaultPyxisApiSecretName      = "pyxis-api-secret"
-	defaultPyxisApiSecretKeyName   = "pyxis_api_key"
+	defaultKubeconfigSecretName        = "kubeconfig"
+	defaultKubeconfigSecretKeyName     = "kubeconfig"
+	kubeConfigSecretAvailable          = "KubeConfigSecretAvailable"
+	defaultGithubApiSecretName         = "github-api-token"
+	defaultGithubApiSecretKeyName      = "GITHUB_TOKEN"
+	gitHubSecretAvailable              = "GithubSecretAvailable"
+	defaultPyxisApiSecretName          = "pyxis-api-secret"
+	defaultPyxisApiSecretKeyName       = "pyxis_api_key"
+	defaultDockerRegistrySecretName    = "registry-dockerconfig-secret"
+	defaultDockerRegistrySecretKeyName = "docker-username"
+	defaultGithubSSHSecretName         = "github-ssh-credentials"
+	defaultGithubSSHSecretKeyName      = "id_rsa"
 )
 
 // ensureKubeConfigSecret will ensure that the kubeconfig Secret is present and up to date.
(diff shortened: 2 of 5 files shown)


controllers/secret.go
```diff
@@ -181,6 +181,23 @@ func (r *OperatorPipelineReconciler) reconcilePyxisAPISecret(meta metav1.ObjectM
 	return nil // Existing Secret found, do nothing...
 }
 
+// dockerRegistrySecretExists will check if the Docker Registry Secret is present.
+func (r *OperatorPipelineReconciler) secretExists(meta metav1.ObjectMeta, secret_purpose string, secret_name string) bool {
```
samira-barouti, line 185: I believe Sid's PR to refactor secrets could be used here.

controllers/resource.go
```diff
@@ -68,5 +68,8 @@ func (r *OperatorPipelineReconciler) reconcileResources(meta metav1.ObjectMeta)
 		return err
 	}
 
+	r.secretExists(meta, "docker registry", "registry-dockerconfig-secret")
+	r.secretExists(meta, "github ssh", "github-ssh-credentials")
```
acornett21, line 72: We should stick with the existing pattern for these secrets as well. It probably would make sense after the refactor PR lands in main.

acornett21:
For `github-ssh-credentials` we need to insure that `id_rsa` exists and is populated as well. Im not sure if we need/have to do this for each `key` for the docker registry, since it seems that `cmd` the cluster already validates for empties on creation.

controllers/resource.go
bcrochet: What's the point of calling this if we are ignoring the return? Is there some kind of side effect happening in secretExists?

bcrochet (changes requested):

controllers/secret.go
dcurran90: great point. will wait for that PR and refactor

controllers/resource.go
dcurran90: however, i'll wait for the refactor.

dcurran90: resolving because updated with refactored cod

controllers/secret.go
```diff
@@ -78,6 +82,36 @@ func (r *OperatorPipelineReconciler) ensurePyxisAPISecret(meta metav1.ObjectMeta
...
+	if err != nil {
+		log.Error(err, "unable to resolve docker registry credentials secret for %s in %s", meta.Name, meta.Namespace)
+		return err
+	}
+	secretName := defaultDockerRegistrySecretName
+	if operatorPipeline.Spec.DockerRegistrySecretName != "" {
+		secretName = operatorPipeline.Spec.DockerRegistrySecretName
+	}
+	r.ensureSecret(secretName, defaultDockerRegistrySecretKeyName, meta)
+	return nil
```
samira-barouti, line 97: I believe this should be return r.ensureSecret(secretName, defaultGithubSSHSecretKeyName, meta). If ensureSecret throws an error, we don't handle it here.

samira-barouti (changes requested):

controllers/secret.go
dcurran90: so the purpose of the function is just to check if the secret exists and log for the users. The secrets are optional and so it should't block if they didn't create it

controllers/secret.go
```diff
@@ -78,6 +82,36 @@ func (r *OperatorPipelineReconciler) ensurePyxisAPISecret(meta metav1.ObjectMeta
...
+	if err != nil {
+		log.Error(err, "unable to resolve github ssh secret for %s in %s", meta.Name, meta.Namespace)
+		return err
+	}
+	secretName := defaultGithubSSHSecretName
+	if operatorPipeline.Spec.GithubSSHSecretName != "" {
+		secretName = operatorPipeline.Spec.GithubSSHSecretName
+	}
+	r.ensureSecret(secretName, defaultGithubSSHSecretKeyName, meta)
+	return nil
```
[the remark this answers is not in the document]
dcurran90: secrets are optional. only purpose of the function is to check if they exist and log for the users knowledge

controllers/secret.go
samira-barouti: You can handle ignoring the error in the caller function. And it is consistent with what has been done for other secrets.

controllers/secret.go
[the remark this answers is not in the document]
bcrochet: I'm with @samira-barouti on this one. If anything, I'd move the logging from the ensureSecret method up to here. You say the secret is optional, but that is not very clear here. Someone else reading the code would need to go looking into the ensureSecret method to find out what this is doing.

bcrochet (changes requested):

bcrochet:
Am I missing something? What in this has to do with pinning?

bcrochet (approved):
/lgtm

controllers/resource.go
```diff
@@ -60,6 +60,14 @@ func (r *OperatorPipelineReconciler) reconcileResources(ctx context.Context, pip
 		return err
 	}
 
+	if err := r.ensureDockerRegistrySecret(ctx, pipeline.ObjectMeta); err != nil {
+		log.Info("Docker Registry Secret not present or correct. Create it to use a private repository.")
```
samira-barouti, line 64: IMO, using log.Warning might be more appropriate since we are handling an error.

samira-barouti (commented):
A nit

acornett21 (commented):
Non blocking: Since these new types are optional, do we want to make them as such explicitly? And maybe also include them in the `sample` yaml, so at least operator hub might end up looking/feeling complete. Or are we going to go down the documentation route for these?

I'm fine either way, but felt this might be a good place to discuss.

bcrochet (changes requested):
Since we don't have the checks landed yet, you will need to run 'make generate' and 'make manifests'. Probably 'make bundle' too.

dcurran90:
updated with make generate, manifests, bundle run. waiting for review

controllers/resource.go
dcurran90: hmm....looks like our logger doesn't have warning? weird. looking into it

dcurran90: yeah, okay. no such thing as warning or debug in this logger. it has V-levels
https://pkg.go.dev/github.com/go-logr/logr#section-readme

I could put like `log.V(2).Info("Docker Registry secret message")`
but I wouldn't say that's necessarily more clear and it would disappear if verbosity is turned down

acornett21: @dcurran90 is correct, there is no `warn` in the library used by controller runtime. I'm fine with this being an `info` and I think it would be bad user experience to log `error` for something that's optional.

itroyano: We should still propagate the err itself, unless it's a `errors.IsNotFound(err)`.

bcrochet: @itroyano has a good point.

bcrochet (changes requested):
There doesn't need to be 3 commits here. 2 maybe... (the initial changes, then the generated stuff).

dcurran90:
updated to two commits

acornett21 (approved):

bcrochet (approved):
/lgtm

PR #44, opened 2021-11-16
SET owner reference for tekton pipelines and tasks

### LINKED ISSUE #40: Resources created by the operator do not get cleaned up when CR is deleted
## Bug Description

The operator creates various resources (pipelines, subscriptions) but not all of them are owned by the custom resource and end up not getting removed when the OperatorPipeline CR is deleted

## Steps to Reproduce:

1) Created OperatorPipeline CR `oc apply -f config/samples/certification_v1alpha1_operatorpipeline.yaml`
2) Wait for operator to resolve
3) Delete OperatorPipeline CR `oc delete -f config/samples/certification_v1alpha1_operatorpipeline.yaml`
4) View pipelines (or any other created resources) in the UI and see that they still exist

## Expected Result

All resources directly created by the operator get removed 

## Actual Result

Some resources (pipelines specifically) stick around

## Additional Context

Mostly has to do with ownership of the new resources not being set to the OperatorPipeline resource. This will likely not be a standard case because we run yaml from another repo to create most (if not all) of the resources in question.
https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/dependencies.go#L24-L76

acmenezes:
This commit includes the following changes:

- Adds a pointer field on the main reconciler struct to the operator pipeline CR
- Update OperatorPipeline field on main reconciler before reconciling resources
- Uses OperatorPipeline on dependencies.go to create the owner reference and set
it for pipelines and tasks.

Those changes enable background deletion when the CR is deleted by the user.

closes #40 

Observations:

It's still not clear if the subscription should be deleted with the CR deletion since it installs the tekton operator once but may have multiple CRs depending on it. I'm not sure on that point. Glad to hear other thoughts on this.

Also although secrets are used by the CR they are not created or managed by the CR. Which means that it's not owned by the CR. I guess they shouldn't be deleted when deleting the CR. Also would like to hear the team's thoughts on this.

### DIFF
--- a/controllers/dependencies.go
+++ b/controllers/dependencies.go
@@ -7,11 +7,13 @@ import (
 	"path/filepath"
 
 	"github.com/go-git/go-git/v5"
+	certv1alpha1 "github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
 	tekton "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
@@ -20,7 +22,7 @@ const (
 	TASK_MANIFESTS_PATH     = "ansible/roles/operator-pipeline/templates/openshift/tasks"
 )
 
-func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.ObjectMeta) error {
+func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(pipeline *certv1alpha1.OperatorPipeline) error {
 
 	// Cloning operator-pipelines project to retrieve pipelines and tasks
 	// yaml manifests that need to be applied beforehand
@@ -54,12 +56,12 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
 
 				// apply pipeline yaml manifests
 				if path == PIPELINE_MANIFESTS_PATH {
-					if errs := r.applyPipelineManifests(filePath, meta); errs != nil {
+					if errs := r.applyPipelineManifests(filePath, pipeline); errs != nil {
 						return errs
 					}
 					// or apply tasks manifests
 				} else {
-					if errs := r.applyTaskManifests(filePath, meta); errs != nil {
+					if errs := r.applyTaskManifests(filePath, pipeline); errs != nil {
 						return errs
 					}
 				}
@@ -75,7 +77,7 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(meta metav1.O
 	return nil
 }
 
-func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, meta metav1.ObjectMeta) error {
+func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, obj metav1.Object) error {
 
 	b, err := os.ReadFile(fileName)
 	if err != nil {
@@ -90,7 +92,7 @@ func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, met
 		return err
 	}
 
-	pipeline.SetNamespace(meta.Namespace)
+	pipeline.SetNamespace(obj.GetNamespace())
 	err = r.Get(context.Background(), types.NamespacedName{Name: pipeline.Name, Namespace: pipeline.Namespace}, pipeline)
 
 	if len(pipeline.ObjectMeta.UID) > 0 {
@@ -104,7 +106,7 @@ func (r *OperatorPipelineReconciler) applyPipelineManifests(fileName string, met
 		if !errors.IsNotFound(err) {
 			return err
 		}
-
+		controllerutil.SetControllerReference(obj, pipeline, r.Scheme)
 		if err := r.Client.Create(context.Background(), pipeline); err != nil {
 			log.Error(err, fmt.Sprintf("failed to create pipeline resource for file: %s", fileName))
 			return err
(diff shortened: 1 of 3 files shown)


controllers/dependencies.go
```diff
@@ -97,6 +97,18 @@ func (r *OperatorPipelineReconciler) applyManifests(fileName string, Namespace s
...
+	// Setting owner reference on created resources to
+	// allow for background deletion
+	ownerRefs := []metav1.OwnerReference{{
+		APIVersion: r.OperatorPipeline.APIVersion,
+		Kind:       r.OperatorPipeline.Kind,
+		Name:       r.OperatorPipeline.Name,
+		UID:        r.OperatorPipeline.UID,
+	}}
+
+	obj.SetOwnerReferences(ownerRefs)
```
samira-barouti, line 110: I believe this method should be leveraged to set the controller as the owner:
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

samira-barouti (changes requested):

controllers/dependencies.go
acmenezes: Interesting. I'm not sure about that. We can both use the controllerutil or the api machinery one that you see on the code. The struct for ownerRefs has other two optional fields being one a boolean that allows for setting the controller as owner. My question is: wouldn't the controller own the primary resource? Being the secondary ones owned by the primary? Like in the code? It actually works pretty well when the CR is deleted. But I didn't test the other one. I'll check that out. And come back here.

samira-barouti: yep, SetOwnerReference set the `BlockOwnerDeletion` to false, but SetControllerReference set it true. Here is the description of the field: https://github.com/kubernetes/apimachinery/blob/19377c9f105d03037e842e151d4605fef730b073/pkg/apis/meta/v1/types.go#L329

controllers/operatorpipeline_controller.go
```diff
@@ -64,6 +65,10 @@ func (r *OperatorPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
+	// setting pipeline pointer on reconciler struct to allow usage
+	// on subreconciler functions
+	r.OperatorPipeline = pipeline
```
bcrochet, line 70: This seems like a bad idea from a concurrency perspective. I'd have to do some investigation, but first glance seems like a problem.
Scenario: Two reconciles happen back to back. Before the first reconcile is complete, the second occurs and overwrites this field. The first reconcile is now in an inconsistent state, as it was operating on a different CR for half of it's reconcile. 

IIUC, there is only one Reconciler. Now, if you wanted to use a mutex, that would be fine.

controllers/dependencies.go
bcrochet: Also, the controllerutil.SetOwnerReference does this for you. You don't have to handle setting all of the metav1.OwnerReference fields.

acornett21: @samira-barouti Is correct we need to use the method linked. We _also_ need to refactor from top town for this controller to this dependecy file to pass in the `OperatorPipeline` struct and not the `meta` struct. SetControllerReference throws an exception when `meta` is passed in. I was testing this extensively yesterday and it does not work otherwise.

acornett21:
We also need to add an `owns` in the below code for `task` and `pipeline` resources.
https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/193baab1c5c3723ca1537025945953a6184f40b6/controllers/operatorpipeline_controller.go#L79

acornett21:
> > We also need to add an `owns` in the below code for `task` and `pipeline` resources.
> > https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/193baab1c5c3723ca1537025945953a6184f40b6/controllers/operatorpipeline_controller.go#L79
> 
> Hi @acornett21 , actually the owns function that belongs to the builder package on controller runtime is setting the watches and enqueueRequestForOwner on behalf of secondary resources. Same thing applies to the setControllerReferece mentioned by @samira-barouti. Either one or the other should be used. I don't think there is a case to use both of them. The operator-sdk way of doing that is using the setupWithManager function calling newControllerManagedBy to accomplish that. It is, behind the scenes, building a complex controller object that is passed as a runnable to the manager workload and started by the manager. That doesn't solve the owner reference element of it though and therefore may not be exactly the same thing this PR addresses. I'm working on this and will answer each one of those comments today as well as update this PR with a rebased push. Thanks!

In my testing, I needed to do _both_ `owns` and `SetControllerReference`, just doing `owns` yield no affected on the `task/pipeline` resources.

controllers/dependencies.go
acmenezes: So in conclusion: 
1 - SetOwnerReference solves the problem of this PR but...
2 - SetControllerReference solves 2 problems:
  a. the background deletion
  b. mark the secondary resource to signal events
3 - The Owns function needs to work in conjunction with SetControllerReference. That is basically the controller saying you generate events and I'm listening. So if you delete a single pipeline or tasks it also triggers the controller and it reapplies all the resources again. How useful is this in this case is a different discussion.

That said all changes are implemented now.

controllers/operatorpipeline_controller.go
acmenezes: As we discussed today, there is way more behind the scenes than a simple struct loading the data. Workers are quite independent of each other. But since it's a little bit obscure I also agree to go through the most comfortable path here. BTW thanks for the help with the code and all tips on rebasing this.

acmenezes:
> > > We also need to add an `owns` in the below code for `task` and `pipeline` resources.
> > > https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/193baab1c5c3723ca1537025945953a6184f40b6/controllers/operatorpipeline_controller.go#L79
> > 
> > 
> > Hi @acornett21 , actually the owns function that belongs to the builder package on controller runtime is setting the watches and enqueueRequestForOwner on behalf of secondary resources. Same thing applies to the setControllerReferece mentioned by @samira-barouti. Either one or the other should be used. I don't think there is a case to use both of them. The operator-sdk way of doing that is using the setupWithManager function calling newControllerManagedBy to accomplish that. It is, behind the scenes, building a complex controller object that is passed as a runnable to the manager workload and started by the manager. That doesn't solve the owner reference element of it though and therefore may not be exactly the same thing this PR addresses. I'm working on this and will answer each one of those comments today as well as update this PR with a rebased push. Thanks!
> 
> In my testing, I needed to do _both_ `owns` and `SetControllerReference`, just doing `owns` yield no affected on the `task/pipeline` resources. Also, if someone uninstalls the operator, we want those resources cleaned up IMO.

Check the replay on the first comment. Totally agree.

bcrochet (approved):
/lgtm

samira-barouti (approved):
/lgtm

PR #49, opened 2021-11-17
Rename publicly-exposed constants to keep them private

### LINKED ISSUE #47: Some constant are publicly available
## Bug Description
Some constants in the codebase are publicly available. These constants need to be private to limit the risk of exposure.

samira-barouti:
Rename publicly exposed constants to ensure they are not accessible externally

* Fixes #47

### DIFF
--- a/controllers/image.go
+++ b/controllers/image.go
@@ -26,15 +26,15 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

PR #50, opened 2021-11-17
Fix some linter issues

bcrochet:
1. len() of a nil slice is zero. No need to explictly check for nil.
2. cfg was unused in the test suite.


samira-barouti (approved):
/lgtm

PR #54, opened 2021-11-18
CSV enhancement

### LINKED ISSUE #43: Update CSV with real information
## Is your feature request related to a problem? Please describe.

We definitely need an appropriate icon and description. 

## Describe the solution you'd like.

We do not have the icon or description at the moment so part of this effort will be to lock them down.
The preflight run may also provide more required updates so watch those results if they are available or we can just create new issues when those results become available.

itroyano:
This PR addresses the following -
1) Adding an icon, description and links to the CSV for user-friendly output to be seen in operatorhub.io/preview
2) passing Preflight bundle check
3) typos and branding.

Fixes #43 

### DIFF
--- a/config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
@@ -5,6 +5,10 @@ metadata:
     alm-examples: '[]'
     capabilities: Basic Install
     operatorframework.io/suggested-namespace: oco
+    operators.operatorframework.io.bundle.channels: alpha
+    operators.operatorframework.io.bundle.channel.default.v1: alpha
+    support: Red Hat
+    repository: https://github.com/redhat-openshift-ecosystem/operator-certification-operator
   name: operator-certification-operator.v0.0.0
   namespace: placeholder
 spec:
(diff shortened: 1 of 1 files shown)


itroyano:
Preview:

image

acornett21:
Nit: Remove the `new lines` that github is complaining about in a files. Other than that, I see no issues.

itroyano:
> Nit: Remove the `new lines` that github is complaining about in a files. Other than that, I see no issues.

done. thanks!

bundle.Dockerfile
```diff
@@ -39,5 +41,7 @@ LABEL quay.expires-after=${quay_expiration}
 
 # Copy files to locations specified by labels.
 COPY --from=builder /workspace/bundle/manifests /manifests/
-COPY --from=builder /workspace/bundle/metadata /metadata/
+# Working around an issue where 'make bundle' will remove annotations from /metadata/annotations.yaml, causing a failed certification
+COPY --from=builder /workspace/annotationsWA /metadata/
+#COPY --from=builder /workspace/bundle/metadata /metadata/
```
samira-barouti, line 46: can we remove this line?

bundle/metadata/annotations.yaml
```diff
@@ -5,10 +5,14 @@ annotations:
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: operator-certification-operator
   operators.operatorframework.io.bundle.channels.v1: alpha
+  operators.operatorframework.io.bundle.channel.default.v1: alpha
```
samira-barouti, line 8: Do we need this file as well? This is a duplicate of `annotationsWA/annotations.yaml`. Can we keep one of them?

bundle.Dockerfile
itroyano: done.

bundle/metadata/annotations.yaml
itroyano: removed it from PR.   I think it's a separate discussion if the file itself needs to be in the codebase.  @bcrochet has a thread on it.

bcrochet (commented):
Please explain what 'annotationsWA' means? Specifically, the 'WA' part? We need to have a longer discussion about what these things bring to us, and whether we should be doing these things in the first place.

itroyano:
@bcrochet 

> Please explain what 'annotationsWA' means? Specifically, the 'WA' part? We need to have a longer discussion about what these things bring to us, and whether we should be doing these things in the first place.

Preflight validates some annotations which `make bundle` doesn't generate. specifically `operators.operatorframework.io.bundle.channel.default.v1`

Moreover if I have an `annotations.yaml` which I've modified and added this annotation - running `make bundle` again with actually REMOVE the extra annotations.

As a WA, I'm keeping the annotations file static.  it doesn't have replaceable versions or operator-spec related attributes anyways.

bundle.Dockerfile
```diff
@@ -27,6 +27,8 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=operator-certification-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.14.0
+LABEL operators.operatorframework.io.bundle.channels.v1: alpha
```
bcrochet, line 30: Are we sure this does the right thing? Can labels be specified as 'LABEL key: value'? The others are 'LABEL key=value'.

bcrochet (changes requested):
Check on dockerfile

bundle.Dockerfile
```diff
@@ -39,5 +40,6 @@ LABEL quay.expires-after=${quay_expiration}
 
 # Copy files to locations specified by labels.
 COPY --from=builder /workspace/bundle/manifests /manifests/
-COPY --from=builder /workspace/bundle/metadata /metadata/
+# Working around an issue where 'make bundle' will remove annotations from /metadata/annotations.yaml, causing a failed certification
```
bcrochet, line 43: This should no longer be necessary.

bcrochet (changes requested):

bundle.Dockerfile
itroyano: reverted

bundle.Dockerfile
itroyano: reverted

itroyano:
Closing this PR in favor of the up to date consolidated #70

PR #60, opened 2021-11-19
Add Github action to ensure CRD is up to dated if the go types have changed

### LINKED ISSUE #57: Pipeline must fail the build if the CRD has not been updated
## Bug Description
Currently, the pipeline does not check if the CRD has been updated if a new field was added to the go type.

## Expected Result
The pipeline must block the PR if the CRD has not been updated.

samira-barouti:
* Fixes #57

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,5 +16,15 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
+
     - run: sudo rm -rf /usr/local/bin/kustomize
-    - run: make test
+
+    - name: Test
+      run: make test
+
+    - name: Manifests
+      run: make manifests && git diff --exit-code
+
+    - name: Generate
+      run: make generate && git diff --exit-code
+


.github/workflows/ci.yml
```diff
@@ -17,4 +17,10 @@ jobs:
       with:
         fetch-depth: 0
     - run: sudo rm -rf /usr/local/bin/kustomize
+    - run: export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
+    - run: export OS=$(uname | awk '{print tolower($0)}')
+    - run: export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.15.0
+    - run: curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+    - run: chmod +x operator-sdk_${OS}_${ARCH} && sudo mv operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
     - run: make test
+    - run: make bundle && git diff --exit-code
```
bcrochet, line 26: For just the CRD, I would rely on 'make manifests' and 'make generate' and check their exit codes. Then you wouldn't need the other code either. It might also make sense to separate these out into other checks? So basically a 'check manifests' and 'check generated code', so that it's explicit as to what failed.

bcrochet (changes requested):

bcrochet (changes requested):

bcrochet (changes requested):
I think these just need to be part of the upper block.

.github/workflows/ci.yml
```diff
@@ -18,3 +18,18 @@ jobs:
...
+    steps:
+    - name: run make bundle
+      run: |
+        export ARCH=`$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)`
+        export OS=$(uname | awk '{print tolower($0)}')
+        export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.15.0
+
+        curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+        chmod +x operator-sdk_${OS}_${ARCH} && sudo mv operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
+        make bundle && git diff --exit-code
```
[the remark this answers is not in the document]
acornett21: FWIW make bundle runs make manifest, but not make generate.

acornett21 (approved):

.github/workflows/ci.yml
```diff
@@ -18,3 +18,18 @@ jobs:
...
+    steps:
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
+      id: go
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - run: |
+        make manifests && git diff --exit-code
```
bcrochet, line 34: In order for this to work properly, I think the commands need to be in individual run: steps. And, I think these can just go between L19 and L20.

So, it would look like this:

```
- run: sudo rm -rf /usr/local/bin/kustomize
- run: make generate && git diff --exit-code
- run: make manifests && git diff --exit-code
- 1run: make test
```

Then you won't need to duplicate the setup.

bcrochet (changes requested):
Quick change request.

bcrochet (approved):
/lgtm

Issue #68, opened 2021-11-29
Bundle Dockerfile shouldn't run Make


itroyano:
## Bug Description

As discussed in Certification Operator Check-In meeting Nov 29th,  there are two main reasons for running `make bundle` - 

1. Align any CRD spec, RBAC and Sample CR changes into the CSV and bundle manifests.
2. Sync the Deployment (manager) of the Bundle with the appropriate Operator image tag, for testing / Preflight.

It was decided that,

1 - actions for updating bundle spec are up to the Developers working on the code/CRD.
Instead of automating this into bundle generation as part of the build, the developer needs to be notified to make the appropriate changes and align the manifests.

2-  see #67

Issue #67, opened 2021-11-29
Bundle CSV should point to "latest" operator image instead of specific GH sha


itroyano:
## Expected Result

```
command:
                - /manager
                image: quay.io/opdev/operator-certification-operator:latest
```

## Actual Result

```
command:
                - /manager
                image: quay.io/opdev/operator-certification-operator:09f54622a8707ebc3e94aaef9a8781cf9bf6cc6d
```

PR #70, opened 2021-11-30
Consolidated PR for bundle and build changes

itroyano:
Fixes: #43 #67 #68 

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -38,10 +38,7 @@ jobs:
       uses: redhat-actions/buildah-build@v2
       with:
         image: ${{ env.IMAGE_REGISTRY }}/operator-certification-operator-bundle
-        tags: ${{ github.sha }} latest
-        build-args: |
-          quay_expiration=1w
-          release_tag=${{ env.RELEASE_TAG }}-alpha
+        tags: latest
         dockerfiles: |
           ./bundle.Dockerfile
 
--- a/config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
@@ -5,6 +5,10 @@ metadata:
     alm-examples: '[]'
     capabilities: Basic Install
     operatorframework.io/suggested-namespace: oco
+    operators.operatorframework.io.bundle.channel.default.v1: alpha
+    operators.operatorframework.io.bundle.channels: alpha
+    repository: https://github.com/redhat-openshift-ecosystem/operator-certification-operator
+    support: Red Hat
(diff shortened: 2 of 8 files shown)


.github/workflows/build-main.yml
```diff
@@ -39,9 +39,6 @@ jobs:
       with:
         image: ${{ env.IMAGE_REGISTRY }}/operator-certification-operator-bundle
         tags: ${{ github.sha }} latest
```
bcrochet, line 41: Do we even care about the sha build now?

bcrochet (changes requested):
Question about the sha tag.

.github/workflows/build-main.yml
itroyano: It's there mostly for tracking daily development, and allowing someone to create an index pointing specifically to a bundle they'd like to test.
Not sure if that's a common use case though.

itroyano: @acornett21 @yashoza19 @spabba17 is there a use case where Preflight is run on an index that points to a specific bundle (not "latest")?
Or a catalog source created for such an index?

bcrochet: If someone wants to do that, they can push to their own quay.io repo. I don't see the utility of it honestly.

itroyano: in that case I'll get rid of it,  if no objections.

config/manifests/bases/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -37,15 +65,17 @@ spec:
...
-  - name: Operator Certification Operator
-    url: https://github.com/redhat-openshift-ecosystem/operator-certification-operator
+  - name: Operator Certification CI Pipeline
+    url: https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/ci-pipeline.md
+  - name: Red Hat OpenShift Pipelines
+    url: https://github.com/openshift/tektoncd-operator
   maintainers:
-  - email: jomckenz@redhat.com
-    name: jomkz
+  - email: certification-operator-dev-team@redhat.com
```
itroyano, line 75: pending - need to update this attribute with the actual email @jomkz created.

jomkz: @itroyano I am anticipating that the request for the email alias will be approved and the email we will use will be:

`certification-operator-devel@redhat.com`

itroyano: updated

bcrochet (approved):
/lgtm

PR #74, opened 2021-11-30
changing kubeconfigSecretName to camelcase, so it passes the CRD validation

acornett21:
- Making it so that the sample file in the operator can be applied directly to a cluster without any changes, since our CRD validates these fields and expects camelCase.


jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #76, opened 2021-12-01
ADD openshift-pipelines operator as a dependency on bundle

acmenezes:
This commit includes:
- removing the subscription and related pipeline parts
- adding a dependency.yaml file with openshift-pipelines operator
- some automated changes made by make generate, manifests and bundle commands

Observation: I need somebody else testing this against a clean cluster other than myself before merging. I did clean my cluster including the tekton CRDs but I'm not 100% sure if I did correctly.


bundle.Dockerfile
```diff
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=operator-certification-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.14.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.0+git
```
bcrochet, line 10: Not sure why this keeps happening. Obviously we want to stick with 1.14. I see this locally as well.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -23,11 +23,11 @@ metadata:
     operatorframework.io/suggested-namespace: oco
     operators.operatorframework.io.bundle.channel.default.v1: alpha
     operators.operatorframework.io.bundle.channels: alpha
-    operators.operatorframework.io/builder: operator-sdk-v1.14.0
+    operators.operatorframework.io/builder: operator-sdk-v1.13.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/redhat-openshift-ecosystem/operator-certification-operator
     support: Red Hat
-  name: operator-certification-operator.v0.0.0-alpha
```
bcrochet, line 30: Do you need to maybe rebase? This shouldn't be changing.

bundle/metadata/dependencies.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
+  - type: olm.package
+    value:
+      packageName: redhat-openshift-pipelines
+      version: "1.5.2"
```
bcrochet, line 5: This needs to be a range. '>=1.5.2' should work. Or, we can use an olm.gvk type of dependency if we don't necessarily want to specify the version.

Also, needs a newline at the end. :)

bcrochet (changes requested):
Couple of requests.

bundle/metadata/dependencies.yaml
acornett21: If GVK type of dependency is available, I vote we go that route, since I don't really think the version matters to us. If for some reason there was a breaking change in a version of the pipelines-operator, I would assume that the team making the tekton tasks would take care of that, and we can be agnostic to version. Just my thoughts.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
acmenezes: This change automatically with make bundle probably. I didn't touch it. Not sure if I should change it manually.

bundle.Dockerfile
acmenezes: Another automatic change. We may need to align operator-sdk versions I believe.

bundle/metadata/dependencies.yaml
acmenezes: I had the impression the that GVK is not fully supported. But since @bcrochet found the recipe for that I'll use that one.

bundle.Dockerfile
acmenezes: Change was made this code snippet is outdated now.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
acmenezes: It seems that we should let this one as it is? Let me know if I need to change it.

bundle/metadata/dependencies.yaml
acmenezes: This one is already changed to @bcrochet suggestion. Let me know if it works now.

acmenezes:
My change actually includes this 2 markers in the zz_generated.deepcopy.go file. No make command can undo this. It seems that I need to change it manually. Even after rebasing.

//go:build !ignore_autogenerated
// +build !ignore_autogenerated

Not sure why this doesn't appear as a change in the files.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
bcrochet: This is fine.

acornett21:
The sample yaml file needs to have the pipeline version removed.

acmenezes:
> 
@acornett21 that's right. Just did it.

bcrochet:
Superseded by #90

Issue #22, opened 2021-11-08
Add in the Ability for the Operator to Update OperatorPipelineStatus


acornett21:
## Is your feature request related to a problem? Please describe.
Currently the operator only logs errors, and unless a cluster admin is tailing the controller manager logs, the `OperatorPipelineStatus`  isn't reflected properly in the `OperatorPipeline` CR. 

## Describe the solution you'd like.
It would be great if the operator could implement OperatorPipelineStatus for any information (secrets/kubeconfigs/etc) that the cluster admin has not provided, so `OperatorPipeline` status is clearly defined.

## Describe alternatives you've considered.
N/A

## Additional context.
N/A

itroyano:
just a quick +1 on this.

PR #77, opened 2021-12-01
implementing a common method to set the StatusCondition

acornett21:
- Relates: #22 
- Implementing a common method to set the StatusCondition on our CR for all resources we reconcile.
- Implementing the above method in ensureGitHubAPISecret.
- To keep PR's easier to read there will be follow up PR's to set StatusCondition for other reconcilers. 


controllers/secret.go
```diff
@@ -55,13 +56,37 @@ func (r *OperatorPipelineReconciler) ensureGitHubAPISecret(ctx context.Context,
 	operatorPipeline, err := r.getPipeline(ctx, meta)
 	if err != nil {
 		log.Error(err, "unable to resolve github secret for %s/%s", meta.Namespace, meta.Name)
+		if err := r.updateStatusCondition(ctx, operatorPipeline, gitHubSecretAvailable, metav1.ConditionFalse, reconcileFailed,
+			err.Error()); err != nil {
+			return err
```
samira-barouti, line 61: I believe the logging is missing when that the status update fails. It has to be done either in here, or in `updateStatusCondition` method.

acornett21: I thought logging the condition update failure might be too much noise for a user in the controller. If an error happens in the update, the reconcile will be triggered again, and hopefully the condition update would happen again. If others agree we should log it, I'd prefer to go with a `WARN` for the level, since condition update is more of a `user friendly` feature and not a `must function` for the operator to work, type of thing.

bcrochet: I agree with @acornett21 . I don't think this necessitates a log message.

bcrochet (approved):
/lgtm

controllers/secret.go
acornett21: Going to resolve this comment, since this is more of a `user experience` nicety. Thanks everyone for the feedback.

jomkz (approved):

PR #85, opened 2021-12-02
update to go/ci/make to see if CI will pass

acornett21:
- title says it all


acornett21:
@bcrochet Look at this craziness. No clue if this is the correct way to go or not, but it works now.

bcrochet (changes requested):
Change Dockerfile then this is good to go. I'm not surprised that there is a difference.

bcrochet (approved):
/lgtm

Issue #86, opened 2021-12-03
Update Docker File and/or Project Dependencies Such that the `as builder` image is not GB of data 


acornett21:
## Is your feature request related to a problem? Please describe.
Currently when trying to build the container image locally it takes a very log time and alot of space (see additional context below). This means that not all developers are able to build this image and test their work in an iterative way.

## Describe the solution you'd like.
I would like the `as builder` image to be as small as possible so that it can be built with a standard `podman vm`. I'm not sure if this is due to the docker file and build process, or the dependencies that the project is pulling in. Further investigation is needed.

## Describe alternatives you've considered.
N/A

## Additional context.
Below are the image sizes of the entire build process, the image marked as `<none>` is the `as builder` image.

```shell
8:16
─acornett at acornett-mac in ~/go/src/github.com/acornett21/operator-certification-operator on fix_weird_build_issue✘✘✘
╰─± podman images
REPOSITORY                                     TAG         IMAGE ID      CREATED        SIZE
quay.io/opdev/operator-certification-operator  latest      b91df0adf025  2 minutes ago  163 MB
<none>                                         <none>      839a40940b8c  3 minutes ago  3.59 GB
docker.io/library/golang                       1.17        e5993351274b  4 hours ago    963 MB
registry.access.redhat.com/ubi8/ubi-minimal    latest      cc54f67dc341  4 weeks ago    107 MB
```

PR #98, opened 2021-12-08
reducing the size of the builder image via go-build time module download

itroyano:
Fixes: #86 

Since we are running `generate fmt vet` followed by `go build` - if we remove `go mod download` and instead run the build itself with `-a`,  we should be able to reduce the size of the builder image from ~3.5GB to ~2.1GB.
another optimization is to copy only the code that we need for compilation as recommended by Operator SDK,  instead of the entire codebase.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,21 @@ FROM docker.io/library/golang:1.17 as builder
 
 ARG release_tag
 
-# Copy the go source
-COPY . /workspace
-
 WORKDIR /workspace
 
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+
+# Copy the code
+COPY main.go main.go
+COPY Makefile Makefile
+COPY hack/ hack/
+COPY api/ api/
+COPY controllers/ controllers/
+
+# Copy git repo for sha info
+COPY .git .git
 
 # Build
 RUN make build RELEASE_TAG=${release_tag}
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 ##@ Build
 
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager -ldflags "-X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.commit=$(SHAVERSION) -X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.version=$(RELEASE_TAG)" main.go
+	go build -a -o bin/manager -ldflags "-X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.commit=$(SHAVERSION) -X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.version=$(RELEASE_TAG)" main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go


Dockerfile
```diff
@@ -6,14 +6,18 @@ FROM docker.io/library/golang:1.17 as builder
...
 
-# Copy the go source
-COPY . /workspace
-
 WORKDIR /workspace
 
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+# Copy the Go Modules manifests
```
bcrochet, line 11: All this is doing is making more layers. The COPY . /workspace isn't that big... Shouldn't we be able to just get away with removing the RUN go mod download?

Makefile
```diff
@@ -99,7 +99,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 ##@ Build
 
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager -ldflags "-X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.commit=$(SHAVERSION) -X github.com/redhat-openshift-ecosystem/operator-certification-operator/version.version=$(RELEASE_TAG)" main.go
```
bcrochet, line 102: Why remove the SHAVERSION? I think that's still very useful, to know where a binary was built from. Not useful on the images, but definitely useful on the binary IMHO

Makefile
```diff
@@ -221,10 +221,13 @@ catalog-push: ## Push a catalog image.
...
 .PHONY: test-sanity
-test-sanity: tidy vet fmt
+test-sanity: tidy git-diff vet fmt
 
 .PHONY: tidy
 tidy:
 	go mod tidy
+
+.PHONY: git-diff
+git-diff:
```
bcrochet, line 231: What's the reason for splitting this out? This is mainly used in CI (if we aren't we should be). So the git diff --exit-code should still be a part of the target.

bcrochet (changes requested):
Couple of asks...

Dockerfile
acornett21: I believe on an `operator-sdk init` this is what you end up getting. ie only copy what you need, this more closely follows that. those layers get throw out though right? since this is the builder image and not the final image of the controller.

Makefile
acornett21: +1

Dockerfile
itroyano: In addition to what @acornett21 said, the codebase over time may have more things that aren't related to the compilation. E.g. more manifests and descriptors.
And obviously this helps filter out large debris like binaries and index img dbs, even if those somehow make it into git.

Makefile
itroyano: Calculating SHAVERSION requires copying .git as well into the builder. Should we?

Makefile
itroyano: IMO it makes sense to separate tidy and git diff into 2 targets.   I took care to call both from `sanity-test` (the only place where tidy was being called).

To me it felt out of place to do a git diff inside a target that's supposed to do go mod tidy.

Originally did that because I wanted to run `tidy` as part of the build & the git diff failed.  but the result was a slightly bigger image at ±2.3gb so went with `-a` eventually which brings size down to 2.1gb.

Dockerfile
bcrochet: Ok. Thanks for the clarification.

Makefile
bcrochet: That tidy target really is just meant for CI consumption IMO. I don't feel strongly about it... But it might end up being more verbose if we need to run 'make tidy git-diff' and 'make fmt git-diff' and 'make vet git-diff'.

bcrochet: And... I don't think tidy should be automatic. Just like the discussion we had about make bundle. CI should inform the change. But that shouldn't be something automatic.

Makefile
itroyano: added.

Makefile
itroyano: But I don't feel strongly against changing it back to what it was.

bcrochet: Maybe then leave it as is since you aren't using it for what you intended and we can have another discussion about it?

bcrochet (changes requested):

Makefile
itroyano: fair enough.  changing back.

bcrochet (approved):
/lgtm

acornett21 (approved):

jomkz (approved):

Issue #94, opened 2021-12-08
Suppress Errors  for Resources Which are Optional 


acornett21:
## Is your feature request related to a problem? Please describe.
When creating a CR, seeing errors for optional resources is confusing.

## Describe the solution you'd like.
I'd like all errors for optional resources to be suppressed, for now this would be
- `registry-dockerconfig-secret`
- `github-ssh-credentials`

I think we should only check for the secret in the cluster if the user has provided a name in the CR. IE [this code](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/secret.go#L181-L185) should be changed. As well as any other optional checks.

```go

	if operatorPipeline.Spec.GithubSSHSecretName != "" {
		if err := r.ensureSecret(ctx, operatorPipeline.Spec.GithubSSHSecretName, defaultGithubSSHSecretKeyName, meta); err !=nil {
			return err
		}
	}
	
	return nil

```

We could then get rid of the default constants.

## Describe alternatives you've considered.
N/A

## Additional context.
Confusing stack trace.

```
2021-12-08T15:06:46.701Z	ERROR	controller_operatorpipeline	could not find existing secret oco/registry-dockerconfig-secret	{"error": "could not find existing secret"}
github.com/redhat-openshift-ecosystem/operator-certification-operator/controllers.(*OperatorPipelineReconciler).ensureSecret
	/workspace/controllers/secret.go:193
github.com/redhat-openshift-ecosystem/operator-certification-operator/controllers.(*OperatorPipelineReconciler).ensureDockerRegistrySecret
	/workspace/controllers/secret.go:171
github.com/redhat-openshift-ecosystem/operator-certification-operator/controllers.(*OperatorPipelineReconciler).reconcileResources
	/workspace/controllers/resource.go:67
github.com/redhat-openshift-ecosystem/operator-certification-operator/controllers.(*OperatorPipelineReconciler).Reconcile
	/workspace/controllers/operatorpipeline_controller.go:66
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:298
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
[...]

PR #99, opened 2021-12-08
update code to only look for optional secrets if a user passes them into the CR

acornett21:
- Fixes: #94 
- update code to only look for optional secrets if a user passes them into the CR
- also fixing docker key name since OC converts the map to a single json key


bcrochet (approved):
/lgtm

jomkz (approved):

PR #102, opened 2021-12-08
Remove redundant secret fetch

### LINKED ISSUE #101: Remove duplicate `Get` code in `ensureSecret` method
## Bug Description
There is duplicate code, where we are calling out to the cluster twice to find a given resource.
Code can be found [here](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/secret.go#L192-L201).

## Version and Command Invocation

Image is `image: quay.io/opdev/operator-certification-operator-index:latest`
which points to latests of the `bundle` image.

## Steps to Reproduce:
1) Run the code
2) See the call to the cluster happening twice for the same resource.

## Expected Result
That we only call the cluster once to `get` a given resource.

## Actual Result
We call for an given secret resource twice.

## Additional Context
N/A

bcrochet:
Also de-export IsObjectFound and FetchObject.

Fixes #101

### DIFF
--- a/controllers/image.go
+++ b/controllers/image.go
@@ -46,7 +46,7 @@ func (r *OperatorPipelineReconciler) reconcileCertifiedImageStream(ctx context.C
 	}
 
 	stream := newImageStream(key)
-	if IsObjectFound(ctx, r.Client, key, stream) {
+	if isObjectFound(ctx, r.Client, key, stream) {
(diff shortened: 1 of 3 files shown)


jomkz (approved):
/lgtm

acornett21 (approved):

PR #103, opened 2021-12-08
Add vars for parts of the image spec

### LINKED ISSUE #97: Makefile should be more flexible with image naming
As a developer, I should just be able to pass IMAGE_REPO=bcrochet, and it would become quay.io/bcrochet/imagename:latest... Instead of IMAGE_TAG_BASE=quay.io/bcrochet/imagename

bcrochet:
Make the image name more flexible by introducing IMAGE_REGISTRY, IMAGE_REPO,
and IMAGE_NAME vars to the Makefile. So, if the only thing that changes
is the repository portion, only IMAGE_REPO=myrepo make docker-build would
be necessary, instead of having to override the entire IMAGE_TAG_BASE.

Fixes #97

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ VERSION="latest"
 RELEASE_TAG ?= "0.0.0"
 CHANNELS="alpha"
 DEFAULT_CHANNEL="alpha"
+IMAGE_REGISTRY ?= "quay.io"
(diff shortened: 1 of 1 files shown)


jomkz (approved):
/lgtm

acornett21 (approved):

Issue #105, opened 2021-12-09
Conditionally Apply Various Pipelines via the CR


acornett21:
## Is your feature request related to a problem? Please describe.
Right now we are applying all of the pipelines `ci/hosted/release` this could be confusing to partners since they only care about the `ci` pipeline.

## Describe the solution you'd like.
Update the CRD to have a boolean for for each for each pipeline type, something like the below.

- `applyCiPipeline: true` - true by default in our sample
- `applyHostedPipeline: false` false by default in our sample
- `applyReleasePipeline: false` false by default in our sample

Based on what the user provides when they apply the CR, we would only apply the  tekton `Pipeline`resource(s) that the user requested. To keep things simple for the operator we can still apply all the tasks so we do not have to worry about looking in the pipeline `yaml` files to distinguish which task(s) goes with which pipeline(s).

We would then check for these booleans in the `applyManifests` code found [here](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/dependencies.go#L110).

## Describe alternatives you've considered.
Open to other ideas, but these are my initial thoughts.

## Additional context.
N/A

itroyano:
The list of pipelines and/or their names can change in the future,  so we can instead of `boolean` have an `enum`?

PR #111, opened 2022-01-11
adding in new flags to the OperatorPipelineSpec to determine what pipeline to install into the cluster when the CR is applied

acornett21:
- Fixes: #105 
- updated OperatorPipelineSpec to include boolean for each type of pipeline.
- added in xDescriptor markers so UI has a true/false toggle for easy usability of new boolean fields.
- added in a `deleteManifests` method to remove a given pipeline if a user changes the value from `true -> false` after a CR is applied.

Below are images of the UI
image

image

### DIFF
--- a/bundle/manifests/certification.redhat.com_operatorpipelines.yaml
+++ b/bundle/manifests/certification.redhat.com_operatorpipelines.yaml
@@ -34,6 +34,18 @@ spec:
           spec:
             description: OperatorPipelineSpec defines the desired state of OperatorPipeline
             properties:
+              applyCIPipeline:
+                description: ApplyCIPipeline determines whether to install the ci
+                  pipeline.
+                type: boolean
+              applyHostedPipeline:
+                description: ApplyHostedPipeline determines whether to install the
+                  hosted pipeline.
+                type: boolean
+              applyReleasePipeline:
+                description: ApplyReleasePipeline determines whether to install the
+                  release pipeline.
+                type: boolean
               dockerRegistrySecretName:
                 description: The name of the secret containing the docker registry
                   credentials secret expected by the pipeline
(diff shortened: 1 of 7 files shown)


bcrochet (changes requested):
Change to be more explicit as we discussed.

acornett21:
@bcrochet I started working on a larger refactor for this like we talked about! Thanks for making this as such. (I'm powerless over here with labels)

bundle/manifests/certification.redhat.com_operatorpipelines.yaml
```diff
@@ -58,6 +70,10 @@ spec:
                 description: The name of the secret containing the pyxis api secret
                   expected by the pipeline
                 type: string
+            required:
+            - applyCIPipeline
+            - applyHostedPipeline
+            - applyReleasePipeline
```
samira-barouti, line 76: You can use the `oneOf` keyword here, and have your CRD only contains one pipeline mode. This will ensure that you don't end up with CRDs whose two modes are set.

samira-barouti (changes requested):

bundle/manifests/certification.redhat.com_operatorpipelines.yaml
acornett21: The user can select multiple pipelines, specifically if it for a RH internal cluster running the certification process. That cluster would need both the `hosted` and `release` pipeline. We wanted to do some sort of `depends on` but there doesn't seem a really good way to do that with descriptors for the UI. We might be able to do something more with the UI around `advanced` features, but that's just look and feel and the CR would still have these values as required. I hope this is clear, but maybe there is a better path forward that we are not seeing.

bcrochet (approved):
/lgtm

jomkz (approved):

PR #114, opened 2022-01-14
Added Get() to updateStatusCondition that stops the application from …

### LINKED ISSUE #112: Controller Randomly Throws `panic`
## Bug Description
Depending on the order that kubernetes processed an `edit` or a `delete` we can sometimes get a panic, since we are not always doing a `get` for the `OperatorPipeline` before we try to `update` the CR.

## Version and Command Invocation
`main branch`

## Steps to Reproduce:
I was only able to reproduce this once here were the steps

1) `kubectl apply -f config/samples/certification_v1alpha1_operatorpipeline.yaml `
2) ` kubectl edit operatorpipelines.certification.redhat.com operatorpipeline-sample` and edit an attribute on the CR
3) immediately after the edit `kubectl delete -f config/samples/certification_v1alpha1_operatorpipeline.yaml`

## Expected Result
I'd expect the controller to process both the edit and delete successfully.

## Actual Result
The delete was processed successfully, then the edit threw a panic, since the CR no longer existed in the cluster. The main code that I see being the issue is the [updateStatusCondition](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/resource.go#L87). We probably need to do a `get` ie `r.Client.Get(ctx, req.NamespacedName, pipeline)`  before we try to do any other actions in the code and short circuit the method if the error returned is an `errors.IsNotFound(err)` error. 

## Additional Context
n/a

spabba17:
Added Get() to [updateStatusCondition](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/controllers/resource.go#L87) that will be executed before any other action in the code and this method is short circuited if the error is not found. This stops the application from throwing a panic.

Fixes #112

### DIFF
--- a/controllers/resource.go
+++ b/controllers/resource.go
@@ -86,6 +86,9 @@ func (r *OperatorPipelineReconciler) reconcileResources(ctx context.Context, pip
 // updateStatusCondition updates the status/reason/message for each condition for which we reconcile
 func (r *OperatorPipelineReconciler) updateStatusCondition(ctx context.Context, pipeline *certv1alpha1.OperatorPipeline,
 	conditionType string, status metav1.ConditionStatus, reason, message string) error {
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: pipeline.GetName(), Namespace: pipeline.GetNamespace()}, pipeline); err != nil {
+		return err
+	}
 	// checking to see if a given condition type is present, since on each reconcile we do not want to set condition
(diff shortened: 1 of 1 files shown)


bcrochet (changes requested):
Can we get a bit more context in the commit message? Thanks.

jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #115, opened 2022-01-17
Refactor reconcilers to be a first-class struct

bcrochet:
Instead of implementing the reconcilers for each resource as a function,
this instead makes them structs that follow the Reconciler interface,
defined in the internal/reconcilers package.

It also makes the status update be its own reconciler, rather than sprinkling
status updates (which could cause unwanted reconcile failures) all throughout.


acornett21:
When you test, can you send the outputted `json` of the CR? I want to make sure we aren't changing what the UI is showing.

bcrochet:
> When you test, can you send the outputted `json` of the CR? I want to make sure we aren't changing what the UI is showing.

It has changed slightly. I think the information is still conveyed properly.

```
apiVersion: certification.redhat.com/v1alpha1
kind: OperatorPipeline
metadata:
  creationTimestamp: "2022-01-17T18:04:18Z"
  generation: 1
  name: example
  namespace: default
  resourceVersion: "8405609"
  uid: 53940161-617b-45ce-aa52-00a4c9f5578c
spec:
  gitHubSecretName: github-api-token
  kubeconfigSecretName: kubeconfig
  pyxisSecretName: pyxis-api-secret
status:
  conditions:
  - lastTransitionTime: "2022-01-17T18:04:21Z"
    message: ""
    reason: ReconcileSucceeded
    status: "True"
    type: PipelineDependenciesAvailable
  - lastTransitionTime: "2022-01-17T18:06:06Z"
    message: ""
    reason: ReconcileSucceeded
    status: "True"
    type: KubeConfigSecretAvailable
  - lastTransitionTime: "2022-01-17T18:06:06Z"
    message: ""
    reason: ReconcileSucceeded
    status: "True"
    type: GithubAPISecretAvailable
  - lastTransitionTime: "2022-01-17T18:06:06Z"
    message: ""
    reason: ReconcileSucceeded
    status: "True"
    type: PyxisApiSecretAvailable
  - lastTransitionTime: "2022-01-17T18:30:23Z"
    message: Marketplace ImageStream found
    observedGeneration: 1
    reason: AsExpected
    status: "True"
    type: Ready
  - lastTransitionTime: "2022-01-17T18:32:39Z"
    message: Certified ImageStream found
    observedGeneration: 1
    reason: AsExpected
    status: "True"
    type: CertifiedIndexReady
  - lastTransitionTime: "2022-01-17T18:33:32Z"
    message: Marketplace ImageStream found
    observedGeneration: 1
    reason: AsExpected
    status: "True"
    type: MarketplaceIndexReady
  observedGeneration: 1
```

internal/reconcilers/certified_image_stream.go
```diff
@@ -0,0 +1,91 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	certifiedIndex                = "certified-operator-index"
+	certifiedImageStreamAvailable = "CertifiedImageStreamAvailable"
```
acornett21, line 19: I think these `Available` constants aren't used anymore from what I can tell.

internal/reconcilers/marketplace_image_stream.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/redhat-openshift-ecosystem/operator-certification-operator/internal/objects"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	marketplaceIndex                = "redhat-marketplace-index"
+	marketplaceImageStreamAvailable = "MarketplaceImageStreamAvailable"
```
acornett21, line 18: Same here.

acornett21 (changes requested):
I think this refactor makes sense to me. It's more clear than handling errors inline and updating the status in the various reconcilers. Also with this, once other `reconcile` functions are migrated to this patter we would not have to worry as much about `errors` and how they may or may not have been propagated back to the main reconcile function.

internal/reconcilers/status.go
```diff
@@ -0,0 +1,141 @@
...
+
+	certifiedImageStream := &imagev1.ImageStream{}
+	err := r.Client.Get(ctx, types.NamespacedName{Namespace: pipeline.Namespace, Name: certifiedIndex}, certifiedImageStream)
+	if err != nil && !errors.IsNotFound(err) {
+		log.WithValues("imagestream", types.NamespacedName{Namespace: pipeline.Namespace, Name: certifiedIndex}).
+			Error(err, "failed to get object")
+		return true, err
+	}
+
+	if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
bcrochet: From what I can tell, it handles the nil case. But it doesn't hurt to add the err != nil. I'll add that shortly.

controllers/operatorpipeline_controller.go
```diff
@@ -63,12 +64,29 @@ func (r *OperatorPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	reconcilers := []reconcilers.Reconciler{
+		reconcilers.NewCertifiedImageStreamReconciler(r.Client, reqLogger, r.Scheme),
+		reconcilers.NewMarketplaceImageStreamReconciler(r.Client, reqLogger, r.Scheme),
+		reconcilers.NewStatusReconciler(r.Client, reqLogger, r.Scheme),
+	}
+
+	requeueResult := false
+	pipeline := currentPipeline.DeepCopy()
+	for _, r := range reconcilers {
+		requeue, err := r.Reconcile(ctx, pipeline)
```
itroyano, line 81: if/when the reconcilers list gets longer in the future, we could use Goroutines here to launch them.

acornett21 (approved):

jomkz (approved):
/lgtm

PR #116, opened 2022-01-20
Move secrets to status reconciler

bcrochet:
Since we are just checking that the secrets exist and are in a correct
key, this is just a status update.


internal/reconcilers/status.go
```diff
@@ -0,0 +1,235 @@
...
+		r.conditionStatus(true),
+		"AsExpected",
+		fmt.Sprintf("%s with name %s found", indexType, indexName),
+		readyCondition))
+
+	return false, nil
+}
+
+func (r *StatusReconciler) reconcileSecretStatus(ctx context.Context, pipeline *certv1alpha1.OperatorPipeline, secretType, secretName, secretKey string) (bool, error) {
+	// Kubeconfig Secret
```
acornett21, line 175: Am I missing something? This isn't just for the kubeconfig right?

internal/reconcilers/status.go
```diff
@@ -0,0 +1,235 @@
...
+	readyCondition := v1.Condition{
+		Type:               fmt.Sprintf("%sReady", secretType),
+		ObservedGeneration: pipeline.Generation,
+		Status:             v1.ConditionUnknown,
+	}
+	log := r.Log.WithValues("status.observedGeneration", pipeline.Generation)
+	secret := &corev1.Secret{}
+	err := r.Client.Get(ctx, types.NamespacedName{Namespace: pipeline.Namespace, Name: secretName}, secret)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.WithValues("kubeconfigsecret", types.NamespacedName{Namespace: pipeline.Namespace, Name: secretName}).
```
acornett21, line 185: Shouldn't this be `secreteType` or `secreteName` here?

internal/reconcilers/status.go
acornett21, line 175: Am I missing something? This isn't just for the kubeconfig right?

internal/reconcilers/status.go
```diff
@@ -0,0 +1,235 @@
...
+			return requeue, err
+		}
+	}
+
+	if len(pipeline.Spec.GithubSSHSecretName) > 0 {
+		requeue, err = r.reconcileSecretStatus(ctx, pipeline, "GithubSSHSecret", pipeline.Spec.GithubSSHSecretName, defaultGithubSSHSecretKeyName)
+		if requeue || err != nil {
+			return requeue, err
+		}
+	}
```
acornett21, line 88: Non Blocking: I know this is the order we were checking before, but maybe it makes sense to move this to after line 99, so this file is easier to read top down.

internal/reconcilers/status.go
bcrochet: Good catch.

internal/reconcilers/status.go
bcrochet: It is not. Missed that on refactor.

internal/reconcilers/status.go
bcrochet: I actually moved it to L71, to put it immediately after the other Github secret.

acornett21 (approved):

jomkz (approved):
/lgtm

PR #117, opened 2022-01-21
Use an emptyDir volume for git repo

bcrochet:
This sets the operator up to only have a single reconcile of the git
repo, then any of the reconcilers (especially the status reconcile)
will be able to use the repo knowing that it's at the latest.

### DIFF
--- a/controllers/dependencies.go
+++ b/controllers/dependencies.go
@@ -40,14 +40,13 @@ func (r *OperatorPipelineReconciler) reconcilePipelineDependencies(ctx context.C
 		return err
 	}
 
-	// creating a tmp director so that each reconcile gets a new directory in case the defer does not execute properly
-	tmpRepoClonePath, _ := os.MkdirTemp("", "operator-pipelines-*")
-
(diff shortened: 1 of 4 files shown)


controllers/dependencies.go
```diff
@@ -190,3 +189,39 @@ func (r *OperatorPipelineReconciler) deleteManifests(ctx context.Context, fileNa
...
+		if err != nil && err != git.NoErrAlreadyUpToDate {
+			log.Error(err, "could not pull remote git repo")
+			return err
+		}
+		ref, err := r.Head()
+		if err != nil {
+			log.Error(err, "could not retrieve current head of")
+			return err
+		}
+		log.Info(fmt.Sprintf("Hash of operator-pipelines HEAD: %s", ref.Hash()))
```
acornett21, line 224: Smart thinking on adding this! We have no idea what we clone today.

acornett21 (approved):

jomkz (approved):
/lgtm

PR #119, opened 2022-03-24
Updating yaml to be in compliance.

dcurran90:
Updated following to pass certification:
--- at beginning of files
line lengths
version numbers


bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -252,6 +329,9 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
+                env:
+                - name: GIT_REPO_PATH
+                  value: /git
                 image: quay.io/opdev/operator-certification-operator:latest
```
acornett21, line 335: This would need to be a `sha` to pass certification. If we have the rbac proxy image somewhere that would need to be a sha as well, and put the proper one in slack.

dcurran90: updated. lmk if you catch anything else

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -252,7 +329,10 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
-                image: quay.io/opdev/operator-certification-operator:latest
+                env:
+                - name: GIT_REPO_PATH
+                  value: /git
+                image: quay.io/opdev/operator-certification-operator@sha256:27645327eff32530c2b7fcbebd9e4fa05290014a2141a2f66b7b5b1edbc7bdb4
```
acornett21, line 335: Since latest gets overwritten, this sha would not always be in quay. Seeing if we should do another release before this PR is merged.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -340,4 +426,4 @@ spec:
   minKubeVersion: 1.21.0
   provider:
     name: Red Hat
-  version: 0.0.0
+  version: 0.0.1
```
acornett21, line 429: Since the container image will be of a `1.0.0` release we should update this version as well.

acornett21 (approved):

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
```diff
@@ -239,7 +316,7 @@ spec:
                 - --upstream=http://127.0.0.1:8080/
                 - --logtostderr=true
                 - --v=10
-                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+                image: registry.redhat.io/openshift4/ose-kube-rbac-proxy@sha256:f000432f07cd187469f0310e3ed9dcf9a5db2be14b8bab9c5293dd1ee8518176
```
bcrochet, line 319: Should pinned names like this be committed? It seems like this should be a part of the release process, and not committed to the repo.

bcrochet (changes requested):
Question about the pinned digests.

bundle/manifests/operator-certification-operator.clusterserviceversion.yaml
acornett21: The release process would then have to go back and commit data after the fact. IE

- Publish an image
- Query the registry for the sha
- Update the make file
- Run make bundle 
- Replace tags with sha's

I personally like having a copy of the bundle in the repo and not just in the `certified-operators` repo.

dcurran90: @bcrochet can we set this as resolved or do you still prefer to remove it?

bcrochet (approved):
/lgtm

PR #123, opened 2022-04-14
Fix the Makefile to use go install for tools

bcrochet:
Go 1.17 deprecated the use of go get for installation of tools. This
switches to use go install to get tool binaries.


jomkz (approved):
/lgtm

acornett21:
@bcrochet Seems like older versions `kustomize` don't support `go install` can you move to `v4@v4.5.4` in this PR?

Issue #128, opened 2022-04-26
operatorPipelinesRelease from the CR is ignored


acornett21:
## Bug Description
We expose the value of `operatorPipelinesRelease` in the CR and this value is ignored when cloning the repo into the container, or on disk when running locally.

## Version and Command Invocation
`main`

## Steps to Reproduce:
1) Run the operator
2) Change `main` in the CR to another branch
3) 

## Expected Result
I would expect the operator checks out the branch specified in the CR.

## Actual Result
The operator always pulls head/main

## Additional Context
n/a

PR #130, opened 2022-04-26
adding in the branchName as part of the git repo clone process

acornett21:
- Fixes: #128 
- Adding in brandName as part of the git repo clone process to make sure we are honoring the value that the user gives us in the CR


bcrochet (approved):
/lgtm

jomkz (approved):

PR #134, opened 2022-05-06
updating openshift annotations to match our minKubeVersion of 1.21.0

acornett21:
- Since our `minKubeVersion` is `1.21.0` we were targeting an incorrect version of OpenShift which our operator would not run on. Updating this to match our bundle in `certified-operators`


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #141, opened 2023-01-17
updating dependencies, operator-sdk versions to release operator

acornett21:
OCO wasn't working on clusters where Pipelines Operator got updated to 1.8.3. When looking into this more, different versions of OpenShift had different versions of Pipelines Operator. This PR aims to address that by switching to a `newest` dependency model, where OLM will apply the newest Pipeliens Operator that it has in a cluster. The pipeline also addresses

- Updating/Migrating Operator SDK to 1.26.0.
- Updating Go Dependencies.
   - Updating Go to 1.19 to satisfy dependency changes. 
- A few files got formatted due to these dependency changes.

Testing:
| OpenShift Version | Operator Pipeline Version | Install Successful |
|------------------- |:-------------------------:|------------------:|
| 4.8                           | 1.5.2                                    | Passed                  | 
| 4.9                           | 1.7.3                                    | Passed                  |
| 4.10                         | 1.8.3                                    | Passed                  |
| 4.11                          | 1.8.3                                   | Not Tested            |  
| 4.12                         | 1.8.3                                    | Passed                  |

**Note: 4.11 was not tested since the previous version was certified here, and has the same dependencies as 4.10 and 4.12**


.github/workflows/ci.yml
```diff
@@ -7,13 +7,13 @@ on:
 jobs:
   sanity:
     name: sanity
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
bcrochet, line 10: Not sure we want to go this far. Could introduce instability without us realizing where it comes from. I'm ok with bumping it. But going flat-out latest I'm not a fan of in CI.

.github/workflows/ci.yml
```diff
@@ -7,13 +7,13 @@ on:
...
     steps:
-    - uses: actions/setup-go@v2
+    - uses: actions/setup-go@v3
       with:
-        go-version: 1.17
+        go-version: 1.19
       id: go
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
bcrochet, line 16: Can we set up the caching like in preflight while we're at it?

Makefile
```diff
@@ -58,7 +58,7 @@ endif
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE):$(RELEASE_TAG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.24
+ENVTEST_K8S_VERSION = 1.25.0
```
bcrochet, line 61: I'm not sure we use this, but is this what the latest operator-sdk would scaffold?

Makefile
```diff
@@ -113,7 +113,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
```
bcrochet, line 116: Is this just a reconciliation with the current scaffold?

Makefile
```diff
@@ -129,6 +129,23 @@ run: manifests generate fmt vet ## Run a controller from your host.
...
+# - able to use docker buildx . More info: https://docs.docker.com/build/buildx/
+# - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+# - be able to push the image for your registry (i.e. if you do not inform a valid value via IMG=<myregistry/image:<tag>> than the export will fail)
+# To properly provided solutions that supports more than one platform you should use this option.
+PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
+.PHONY: docker-buildx
+docker-buildx: test ## Build and push docker image for the manager for cross-platform support
+	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
+	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
+	- docker buildx create --name project-v3-builder
```
bcrochet, line 143: Any references to 'docker' should be ${IMAGE_BUILDER}

Makefile
```diff
@@ -170,9 +187,9 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
 
 ## Tool Versions
-KUSTOMIZE_VERSION ?= v4.5.4
-CONTROLLER_TOOLS_VERSION ?= v0.9.0
-OPERATOR_SDK_VERSION ?= "v1.22.1"
+KUSTOMIZE_VERSION ?= v3.8.7
```
bcrochet, line 190: Why are we going backwards?

bcrochet (changes requested):
A few questions and a change.

Makefile
acornett21: Yeah that's just the latest they can scaffold and is in the upgrade docs. We don't `use` it per say, but other cmds download that binary.

Makefile
acornett21: Yes, just following the upgrade docs.

.github/workflows/ci.yml
acornett21: I was thinking similar, but preflight was using this so I thought it would be okay. It looks like we have two choices.

- `ubuntu-20.04`
- `ubuntu-22.04` (which is `latest`)

based on this depreciation [article](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/). Should we just go to the `20.04` for now?

bcrochet: 22.04 would be fine. Updating it once every two years would not be horrible.

Makefile
bcrochet: Where is LOCALBIN defined? I don't see that in the Makefile

bcrochet (changes requested):
More comments.

Makefile
acornett21: It's been defined [here](https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/Makefile#L162-L164) since the initial commit I think.

.github/workflows/ci.yml
acornett21: I added the caching.

bcrochet (approved):
/lgtm

api/v1alpha1/groupversion_info.go
```diff
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the certification v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=certification.redhat.com
```
komish, line 19: I don't think you want a space here. IIRC, markers are commonly distinguished from comments _because_ they lack this space. I'm not sure if it's enforced or a semantic, though.

komish: Another common example is
```
//go:embed ...
```

bundle/metadata/dependencies.yaml
```diff
@@ -2,4 +2,4 @@ dependencies:
   - type: olm.package
     value:
       packageName: openshift-pipelines-operator-rh
-      version: "1.7.1"
+      version: ">=1.5.2"
```
komish, line 5: We're supporting older versions of openshift-pipelines-operator-rh with this directive, right? Is that the goal?

komish (approved):
Left a few comments, which are mostly nits. As we discussed offline, I don't really consider myself a maintainer of this project, and so I don't know much of its operation. That said, logistically, this patch seems fine. 

To that end.

/lgtm

api/v1alpha1/groupversion_info.go
acornett21: fixed, something was auto formatting this, still don't know what.

bundle/metadata/dependencies.yaml
acornett21: This is a great question. Since we have a hard dependency on the pipelines operator, we need to may sure any version greater then 1.5.2 is in the cluster (since this is the version that has the CRD's we rely on). Ideally, as pipelines operator released new versions we'd update to the latest version, but they do not release all new versions to all OpenShift versions. So to make sure that OCO (this operator) works on all clusters we need to support 4.8 onward, this syntax makes sure we get the newest version in whatever cluster runs this operator. Hope this is clear.

komish: Got it, and the common alternative to this approach is to release multiple streams, each with a hard-pinned dependency vs. this which is more flexible but releases a single stream of this project?

acornett21: Correct. I'm not sure how we'd go to a multi-stream approach with a hard pinned version at this point(in a logistical sense), nor do I think we want that kind of overhead when trying to certify.

bcrochet (approved):
/lgtm

Issue #143, opened 2023-03-07
Add Pipelines SCC and Update Pipeline SA to Reference new SCC


acornett21:
## Is your feature request related to a problem? Please describe.
With versions of tekton operator 1.9 and above, they have removed `privileged escalation` from their `scc`. This means that `operator-pipelines` codebase now has their own `scc`.  Since `operator-pipelines` now has it's own scc, `oco` needs to make sure that this is applied properly. 

## Describe the solution you'd like.
- Apply the new `scc` after we `clone` it from git. (This might happen automatically, depending on which directory the `scc` lives in.)
- Add the `scc` to the `sa` programmatically, probably in it's own sub-reconciler. (see below example)
   - `oc adm policy add-scc-to-user pipelines-custom-scc -z pipeline` 

## Describe alternatives you've considered.
N/A

## Additional context.
Below are the instructions that `operator-pipelines` has listed to accomplish this manually.

```
# Create a new SCC
oc apply -f ansible/roles/operator-pipeline/templates/openshift/openshift-pipelines-custom-scc.yml
# Add SCC to a pipeline service account
oc adm policy add-scc-to-user pipelines-custom-scc -z pipeline
```

PR #144, opened 2023-05-10
adding scc, role, rolebinding logic

acornett21:
- Adds logic to add SCC, ClusterRole, and ClusterRole binding that exist in `operator-pipelines` repo that are needed to support newer version of `tekton` operator.
- Adds `finalizer` to our CR so we can probably remove these cluster scoped resources in the case of multiple `OperatorPipelines` CustomResources
- Fixes: #143


config/rbac/role.yaml
```diff
@@ -61,6 +61,27 @@ rules:
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - '*'
```
komish, line 84: Let's be more explicit with the verbs we need here, if possible.

controllers/operatorpipeline_controller.go
```diff
@@ -18,19 +18,25 @@ package controllers
...
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const operatorPipelineFinalizer = "certification.redhat.com/finalizer"
```
komish, line 38: Should we export this constant so that others can use it (if they need to)?

controllers/operatorpipeline_controller.go
```diff
@@ -64,6 +72,48 @@ func (r *OperatorPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			// if the length is 1 we know this is the last CR in a given namespace and can remove the ClusterRoleBinding
+			// associated with this namespace
+			if len(namespacePipelines.Items) == 1 {
+				if err := r.deleteClusterRoleBinding(ctx, log, currentPipeline.Namespace); err != nil {
+					return ctrl.Result{}, err
+				}
+			}
+
+			clusterPipelines := &certv1alpha1.OperatorPipelineList{}
+			if err := r.Client.List(ctx, clusterPipelines); err != nil {
```
komish, line 95: Just so I'm clear, because this doesn't have ListOptions, we're querying for OperatorPipelineList and the items will span all namespaces?

controllers/operatorpipeline_controller.go
```diff
@@ -85,10 +135,72 @@ func (r *OperatorPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+
+	return nil
+}
+
+func (r *OperatorPipelineReconciler) deleteClusterRoleBinding(ctx context.Context, log logr.Logger, nameSpace string) error {
+	log.Info("starting deletion of owned ClusterRoleBinding")
+
+	listOption := client.MatchingLabels{
+		fmt.Sprintf("%s/%s", reconcilers.OperatorPipelineLabel, "cluster-resource"): "true",
+		fmt.Sprintf("%s/%s", reconcilers.OperatorPipelineLabel, "metadata.name"):    nameSpace,
```
komish, line 187: Totally would have just made these constants, and done the fmt.Sprintf-ing at the constant declaration :laughing:

controllers/operatorpipeline_controller.go
```diff
@@ -97,5 +209,8 @@ func (r *OperatorPipelineReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&imagev1.ImageStream{}).
 		Owns(&tekton.Pipeline{}).
 		Owns(&tekton.Task{}).
+		Owns(&securityv1.SecurityContextConstraints{}).
+		Owns(&rbacv1.ClusterRole{}).
+		Owns(&rbacv1.ClusterRoleBinding{}).
```
komish, line 214: Hrm, do you still need to own these resources? Or do you need `For` instead?

internal/reconcilers/pipeline_dependencies.go
```diff
@@ -82,23 +90,44 @@ func (r *PipelineDependenciesReconciler) Reconcile(ctx context.Context, pipeline
...
+		return true, err
+	}
+
+	// apply the cluster role binding with modifications below
+	tmpFile, err := r.modifyAndSaveTempClusterRoleBinding(ctx, filepath.Join(gitPath, baseManifestsPath, clusterRoleBindingYml), pipeline, new(rbacv1.ClusterRoleBinding))
+	if err != nil {
+		return true, err
+
+	}
+	//todo-adam is this necessary? isn't temp cleaned up or no? probably better safe then sorry?
```
komish, line 113: Leftover comment

internal/reconcilers/pipeline_dependencies.go
```diff
@@ -163,3 +205,52 @@ func (r *PipelineDependenciesReconciler) deleteManifests(ctx context.Context, fi
...
+	log := r.Log.WithName("modifyAndSaveTempClusterRoleBinding")
+
+	b, err := os.ReadFile(fileName)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Couldn't read manifest file for: %s", fileName))
+		return "", err
+	}
+
+	if err = yamlutil.Unmarshal(b, &obj); err != nil {
+		log.Error(err, fmt.Sprintf("Couldn't unmarshall yaml file for: %s", fileName))
```
komish, line 219:
```suggestion
		log.Error(err, fmt.Sprintf("Couldn't unmarshal yaml file for: %s", fileName))
```

internal/reconcilers/pipeline_dependencies.go
```diff
@@ -163,3 +205,52 @@ func (r *PipelineDependenciesReconciler) deleteManifests(ctx context.Context, fi
...
+		return "", err
+	}
+
+	// type asserting to ensure proper kind inorder to update specific values before updating/creating in cluster
+	crb, ok := obj.(*rbacv1.ClusterRoleBinding)
+	if !ok {
+		return "", fmt.Errorf("could not type assert client.Object as ClusterRoleBinding")
```
komish, line 226: Is this function going to be called on non-ClusterRoleBindings? If so, do we want to handle this error differently? In practice, this shouldn't be run on things that aren't the ClusterRoleBinding, so an error here should be a no-op for other object types, no?

komish (changes requested):
Just recommending a couple of changes, nothing major.

internal/reconcilers/pipeline_dependencies.go
acornett21: The comment is valid, I meant to ask others if this is needed or not. :)

What are your thoughts? do we usually cleanup on files put in a `tmp` dir?

internal/reconcilers/pipeline_dependencies.go
acornett21: This method only called for `ClusterRoleBindings` since this is the only file that is templated, ie we need to replace the CRB's `name` and reference `namespace` for the `ServiceAccount` so everything is linked properly. If I don't do a type assertion I can't modify the object how I need to.

controllers/operatorpipeline_controller.go
acornett21: This `owns` is inside a `for` for `OperatorPipeline`, and we are adding an owner to each one of these resources, maybe this diff just doesn't show this properly, but I think this is already how you expect it. Please lmk if you are thinking something else.

controllers/operatorpipeline_controller.go
acornett21: Yes correct, I wasn't sure if querying twice was the best option, but since the `ClusterRoleBinding` is tied to a namespace `ServiceAccount` and references the `namespace` I ended up with this approach, though I'm really open to other ideas if there are any.

Maybe this needs comments on the `List` calls, right now I have the explanation on the `if` logic after these calls. Thoughts?

config/rbac/role.yaml
acornett21: I think we do need `*` otherwise I believe I was getting exceptions, though maybe a list of all of the verbs is better? FWIW other operators that are in the certified repo are using `*` for this. (not that that is saying much though), but thought I'd note that.

controllers/operatorpipeline_controller.go
acornett21: `fmt.Sprintf` can't be uses in a constant declaration apparently, but i just made the entire strings constants.

controllers/operatorpipeline_controller.go
komish: Maybe just a small blurb explaining the intent here (e.g. move the "if this is instance is the last instance cluster-wide, also delete <N>" here)

controllers/operatorpipeline_controller.go
komish: Did you end up getting events when a user deletes the CRB/CR/SCC that's owned by this controller? To remake it?

internal/reconcilers/pipeline_dependencies.go
komish: Can't hurt to clean up. I'd say it's fine.

internal/reconcilers/pipeline_dependencies.go
komish: The type assertion is fine. I'm just unsure if we need the error here, or we should treat it as a no-op and do nothing (i.e. don't emit an error). I suppose if we're handling the error such that "wrong type" is just ignored at call time, then this question doesn't matter.

controllers/operatorpipeline_controller.go
acornett21: I added two more comments, please take a look at all the comments and let me know if they are clear, or if having this many is confusing.

controllers/operatorpipeline_controller.go
acornett21: The reconciler does not get events for the cluster scoped resources, so if they are deleted on accident by a user we can't recreate them.

bcrochet (approved):
/lgtm

controllers/operatorpipeline_controller.go
komish: Just closing the loop - you _can_ recreate them, but you probably don't automatically recreate them on deletion. Just when you go back through your reconcile loop. Neither here nor there, resolving this comment thread.

komish (approved):
/lgtm

PR #145, opened 2023-05-10
adding golangci linting to project and correcting all errors found

acornett21:
- Adding linting to this project
- Followed preflights format/layout
- Fixed any issues that were outputted

### DIFF
--- a/api/v1alpha1/groupversion_info.go
+++ b/api/v1alpha1/groupversion_info.go
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the certification v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=certification.redhat.com
+// +kubebuilder:object:generate=true
+// +groupName=certification.redhat.com
 package v1alpha1
 
 import (
--- a/controllers/operatorpipeline_controller.go
+++ b/controllers/operatorpipeline_controller.go
@@ -19,11 +19,12 @@ package controllers
 import (
 	"context"
 
+	"github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
+	"github.com/redhat-openshift-ecosystem/operator-certification-operator/internal/reconcilers"
+
 	"github.com/go-logr/logr"
 	imagev1 "github.com/openshift/api/image/v1"
 	securityv1 "github.com/openshift/api/security/v1"
-	certv1alpha1 "github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
-	"github.com/redhat-openshift-ecosystem/operator-certification-operator/internal/reconcilers"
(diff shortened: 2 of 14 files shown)


api/v1alpha1/groupversion_info.go
```diff
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the certification v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=certification.redhat.com
+// +kubebuilder:object:generate=true
```
bcrochet, line 18: This doesn't seem right? Do they still work with the space?

controllers/operatorpipeline_controller.go
```diff
@@ -53,7 +54,7 @@ func (r *OperatorPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	reqLogger := logf.FromContext(ctx, "Request.Namespace", req.Namespace, "Request.Name", req.Name)
 	reqLogger.Info("Reconciling OperatorPipeline")
 
-	currentPipeline := &certv1alpha1.OperatorPipeline{}
```
bcrochet, line 56: If we want this to remain certv1alpha1, we can add that to the golangci-lint config.

bcrochet (commented):
Couple of questions.

controllers/operatorpipeline_controller.go
acornett21: I can make that change, first pass was just to get this working.

api/v1alpha1/groupversion_info.go
acornett21: I have no idea what is changing this in _this_ file **only**, but it should not have a space. The other files do not have a space and do not get changed.

controllers/operatorpipeline_controller.go
komish: If you added this to the alias, do you need to rerun golangci-lint to make this certv1alpha1 again?

komish (approved):
/LGTM

There's one question I had on an existing thread. Otherwise, looks fine.

controllers/operatorpipeline_controller.go
acornett21: Good question. The answer is no, `golangci-lint` doesn't actually validate/change a `non alias` to an `alias` for imports that don't have an `alias`. What moving this to the `alias` file does, is that if someone does use an `alias` it will throw an error. For now though, since there isn't any conflict for `alpha1`, no `alias` is needed. 

In reality, we are just reserving this for later use by this name, say for example if we had a file with two `alpha1`'s (from two different libraries/packages) we needed to import, we should `alias` them both and then the linter would throw an error if we did not use `certv1alpha1` for our import. Hope this is clear, though I realize this might be more confusing.

komish: Resolving

PR #147, opened 2023-06-07
Adding a namespace to the example CR metadata

### LINKED ISSUE #146: Set namespace of the example `OperatorPipelines` resource to `oco`
## Bug Description

Since the operator bundle is installed in "all namespaces" mode,  OCO is running on the `openshift-operators` namespace.
If the user clicks `create OperatorPipeline` from here,  the default namespace for this CR is `openshift-operators` too.
Form View UI doesn't allow changing that. only yaml view does.
It's easy to make the mistake here and proceed to create all tasks and pipelines in operators namespace, while the secrets and configs are in `oco` NS per documentation.

putting explicitly
```
namespace: oco
```
in the sample resource of CSV, should resolve this.

image

itroyano:
Enforce creating the example operand in the same namespace that includes the secrets and configs.
Fixes #146

### DIFF
--- a/config/samples/certification_v1alpha1_operatorpipeline.yaml
+++ b/config/samples/certification_v1alpha1_operatorpipeline.yaml
@@ -2,6 +2,7 @@ apiVersion: certification.redhat.com/v1alpha1
 kind: OperatorPipeline
 metadata:
   name: operatorpipeline-sample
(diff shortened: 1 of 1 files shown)


bcrochet (approved):
/lgtm

Issue #150, opened 2023-10-25
Setting OperatorPipelinesRelease does not fetch the version


sxd:
## Bug Description

When setting the `OperatorPipelinesRelease` for OperatorPipeline like this:

```
operatorCertification.OperatorPipeline{
		ObjectMeta: metav1.ObjectMeta{
			Name:      PipelineName,
			Namespace: testingNamespace,
		},
		Spec: operatorCertification.OperatorPipelineSpec{
			ApplyCIPipeline:          true,
			ApplyHostedPipeline:      false,
			ApplyReleasePipeline:     false,
			GitHubSecretName:         githubSecretName,
			KubeconfigSecretName:     kubeconfigSecretName,
			OperatorPipelinesRelease: "v1.0.98",
			PyxisSecretName:          pyxisSecretName,
		},
	}
```

When running this I get this error:
```
status:
  conditions:
  - lastTransitionTime: "2023-10-25T14:05:35Z"
    message: Local repo unavailable
    observedGeneration: 1
    reason: NotFound
    status: "False"
    type: GitRepoReady
  observedGeneration: 1
```
Looks like everything is related to this part of the code https://github.com/redhat-openshift-ecosystem/operator-certification-operator/blob/main/internal/reconcilers/pipeline_git_repo.go#L59 where we only get a branch and not the tag, so we don't get a `Release` but rather we only get a branch.

This is kind of difficult when there's a small change like the one reported here https://github.com/redhat-openshift-ecosystem/operator-pipelines/issues/530 because we should be getting a version to run the certification pipeline

acornett21:
hi @sxd thanks for reporting this issue. We'll take a look at the best way to be able to use a branch or a tag.

sxd:
Hi @acornett21 

We hit this issue again because of this commit https://github.com/redhat-openshift-ecosystem/operator-pipelines/commit/1651f69c9e2a6df14e15a82d3167ec34501ab487 now we can't run anything because we can only use `main` and not the release version.

There's a way I can test your patch? like a development version of the operator I can deploy ?

Regards!

acornett21:
@sxd version `1.1.3` has been released and should be in your clusters catalog to use. Let us know if you have any issues.

sxd:
@acornett21 thank you so much!! I'm going to test it and try it

PR #153, opened 2023-11-07
updating git clone logic to support tags

acornett21:
- Fixes: #150 
- Updating clone logic to support tags
- Switching from `pull` (which only supported branches) to `fetch` and `checkout` by hash

### DIFF
--- a/internal/reconcilers/pipeline_git_repo.go
+++ b/internal/reconcilers/pipeline_git_repo.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/redhat-openshift-ecosystem/operator-certification-operator/api/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/operator-certification-operator/internal/errors"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/storer"
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


internal/reconcilers/pipeline_git_repo.go
```diff
@@ -52,31 +55,65 @@ func (r *PipelineGitRepoReconciler) Reconcile(_ context.Context, pipeline *v1alp
...
+		var err error
+		r, err = git.PlainOpen(targetPath)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	// If r is still both clone and open were unsuccessful, returning err to avoid a later panic
+	if r == nil {
+		return "", errors.New("could not clone or open repo")
```
bcrochet, line 77: Rather than import errors, and have to change all the other errors, use `fmt.Errorf`

internal/reconcilers/pipeline_git_repo.go
```diff
@@ -52,31 +55,65 @@ func (r *PipelineGitRepoReconciler) Reconcile(_ context.Context, pipeline *v1alp
...
-	if err != nil {
+	if r == nil && err == git.ErrRepositoryAlreadyExists {
+		var err error
+		r, err = git.PlainOpen(targetPath)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	// If r is still both clone and open were unsuccessful, returning err to avoid a later panic
```
bcrochet, line 75:
```suggestion
	// If r is nil both clone and open were unsuccessful, returning err to avoid a later panic
```

internal/reconcilers/pipeline_git_repo.go
```diff
@@ -52,31 +55,65 @@ func (r *PipelineGitRepoReconciler) Reconcile(_ context.Context, pipeline *v1alp
...
+		}
+		ref = reference
+		return storer.ErrStop
+	}); err != nil {
 		return "", err
 	}
-	ref, err := r.Head()
+
+	if ref == nil {
+		return "", errors.New("requested release is not in repository")
```
bcrochet, line 104: Same as above about `fmt.Errorf` usage

bcrochet (changes requested):
Couple of changes.

bcrochet (approved):
/lgtm

Issue #152, opened 2023-11-01
Update to Latest Version of Tekton Dependency


acornett21:
## Is your feature request related to a problem? Please describe.
N/A

## Describe the solution you'd like.

Update the dependency to the latest possible version whenever this feature is implemented.

## Additional context.

Tekton has deprecated `v1beta` for a few releases now, so we should update to the latest version after `operator-pipelines` has migrated to `v1` API's

PR #157, opened 2024-01-30
updating tekton dependency and migrating to v1 api's

acornett21:
- Fixes: #152
- Updating to tekton `v1` apis which is now being used in `operator-pipelines`


bcrochet (approved):
/lgtm

PR #158, opened 2024-08-02
updating k8s to 1.30 and go to 1.22.5

acornett21:
- Updating k8s to 1.30 (and any other support dep for compatibility)
- Updating go to 1.22.5
- Updating golangci-lint
   - Had to remove depguard, since it is not compatible issues: https://github.com/golangci/golangci-lint/issues/3906#issuecomment-1591884729
   - `deadcode` changed to `unused`


.golangci.yaml
```diff
@@ -5,8 +5,6 @@ run:
 linters:
   enable:
   - asciicheck
-  - deadcode
```
itroyano, line 8: these got removed by the newer linter version?

Makefile
```diff
@@ -58,7 +58,7 @@ endif
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE):$(RELEASE_TAG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.29.0
+ENVTEST_K8S_VERSION = 1.30.0
```
bcrochet, line 61: For all of these... I'd like to see these turned into ?= assignments so that we can define them outside of the Makefile, to be able to test the versions in CI without having to actually create a PR to do it. Or even have a current/next type of deal so we can be testing both on a rolling basis. Just a though. Not necessarily relevant to this PR. 😄

bcrochet: It looks like this might be the only one that does not have a ?= assignment.

bcrochet (approved):
/lgtm

.golangci.yaml
acornett21: I pointed this out in the notes that `deadcode` is now `unused`

Makefile
acornett21: That does look like the only one. I like the thought of what you are saying, the only caveat  to doing this or that testing, is that most of these dependencies need to be in lock-step with one another to work properly. Sometimes there is variances, but usually not from what I've seen.

PR #159, opened 2024-08-07
adding logic to add owner reference to ImageStream

acornett21:
## Motivation
During some testing, I realized that we were never cleaning up the `ImageStream` CR, when an `OperatorPipeline` was deleted. After digging further, I found we were never setting the owner reference, (also we never really call `Create` for `ImageStream` directly, this is abstracted from since we only `Create` and `ImageStreamImport`).

## Change
Calling `SetControllerReference` which sets the `owner` reference properly, and calling update, when secondary reconciles happen.

### DIFF
--- a/internal/reconcilers/certified_image_stream.go
+++ b/internal/reconcilers/certified_image_stream.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
@@ -44,6 +45,18 @@ func (r *CertifiedImageStreamReconciler) Reconcile(ctx context.Context, pipeline
 	stream := newImageStream(key)
 	if objects.IsObjectFound(ctx, r.Client, key, stream) {
 		log.Info("existing certified image stream found")
+
+		// setting owner reference on ImageStream CR, so CR gets garbage collected on OperatorPipeline deletion.
+		// ignoring error, since we do not need/want to requeue on this failure,
+		// and this should self correct on subsequent reconciles.
+		err := controllerutil.SetControllerReference(pipeline, stream, r.Scheme)
+		if err != nil {
+			log.Info("unable to set owner on certified image stream, "+
+				"this resource will need to be cleaned up manually on uninstall", "error", err.Error())
+			return false, nil
+		}
+		_ = r.Update(ctx, stream)
+
 		return false, nil // Existing ImageStream found, do nothing...
 	}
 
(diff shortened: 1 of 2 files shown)


internal/reconcilers/certified_image_stream.go
```diff
@@ -44,6 +45,13 @@ func (r *CertifiedImageStreamReconciler) Reconcile(ctx context.Context, pipeline
 	stream := newImageStream(key)
 	if objects.IsObjectFound(ctx, r.Client, key, stream) {
 		log.Info("existing certified image stream found")
+
+		// setting owner reference on ImageStream CR, so CR gets garbage collected on OperatorPipeline deletion.
```
skattoju, line 49: Is it worth checking for an owner reference with `HasControllerReference` before attempting to set / update owner references ?

acornett21: I don't think this really buys us anything, since there isn't another controller that could own this. We don't do this check in other reconcilers.

internal/reconcilers/certified_image_stream.go
```diff
@@ -44,6 +45,13 @@ func (r *CertifiedImageStreamReconciler) Reconcile(ctx context.Context, pipeline
...
+		// ignoring error, since we do not need/want to requeue on this failure,
+		// and this should self correct on subsequent reconciles.
+		_ = controllerutil.SetControllerReference(pipeline, stream, r.Scheme)
```
komish, line 52: You need to handle this error. I would log if there's an error, and only update if there's _no_ error. In both cases, skip reconciliation to make this a no-impact operation.

komish (changes requested):
Just 1 change. Otherwise, LGTM.

internal/reconcilers/certified_image_stream.go
acornett21: I've never seen handling this error elsewhere, are you saying we should handle this specifically because of the use case we are trying to address?

komish: Nah. I'm saying that handling this error gives you (the developer) more visibility into what's going wrong without having to add LoC at a later date.

`controllerutil.SetControllerReference` throws an error if applying your controller reference would conflict with an existing reference on the object. If that's happening, I would imagine you'd want to know that without having to write extra LoC just to debug it.

acornett21: I think the latest commit should address this, lmk if other changes are needed.

internal/reconcilers/marketplace_image_stream.go
```diff
@@ -42,7 +43,19 @@ func (r *MarketplaceImageStreamReconciler) Reconcile(ctx context.Context, pipeli
 
 	stream := newImageStream(key)
 	if objects.IsObjectFound(ctx, r.Client, key, stream) {
-		log.Info("existing marketplace image stream found")
+		log.Info("existing certified image stream found")
```
komish, line 46: Typo:
```suggestion
		log.Info("existing martkeplace image stream found")
```

komish (changes requested):
Typo. Otherwise LGTM.

komish (approved):
/lgtm

Issue #154, opened 2023-12-07
Can't pull the right certified-operator-index image in a brand new cluster


sxd:
## Bug Description

When running the operator to certify a new version, we found out that it was pulling the wrong image of the `certified-operator-index` because it was 4.15, which "doesn't exist" yet, leading to a problem like this

```
++ echo registry.redhat.io/redhat/certified-operator-index:v4.15
++ rev
++ cut -f1 -d/
++ rev
+ INDEX_IMG_STREAM=certified-operator-index:v4.15
+ FROM_INDEX=image-registry.openshift-image-registry.svc:5000/oco/certified-operator-index:v4.15
+ podman pull image-registry.openshift-image-registry.svc:5000/oco/certified-operator-index:v4.15
Trying to pull image-registry.openshift-image-registry.svc:5000/oco/certified-operator-index:v4.15...
Error: initializing source docker://image-registry.openshift-image-registry.svc:5000/oco/certified-operator-index:v4.15: reading manifest v4.15 in image-registry.openshift-image-registry.svc:5000/oco/certified-operator-index: name unknown
```
Even if the operator only supports v4.14 as a top  version, the operator certification downloads the the v4.15 tag and the container can't pull the v4.14, to clarify, I don't know if this is the right place to ask for help, but the certification process doesn't work because of this.

The fix is clear, if you set the operator to have the maximum support on 4.15, it works really nice, that's why I open this issue, if it's not the right please, can you point me the right one?

acornett21:
Hi @sxd I'm not sure I understand the question/issue fully, so let me ask some questions...
- what version of OCP are you installing this operator on?
- what version(s) of OCP does the operator under test support?

acornett21:
@sxd Answers to the above would be helpful, but I think I kind of understand. It seems that the `--all` cmd even in the `oc` cli is not important **every** tag into this particular image stream. Since the CLI can't do it, the likelihood of the API calls in our operator doing them is probably not possible, which we'll have to address another way.

As a workaround after the operate is installed and CR reconciled the below should work (replace v4.14 with the version of your cluster):
```
oc import-image certified-operator-index:v4.14 \                  
  --from=registry.redhat.io/redhat/certified-operator-index:v4.14 \
  --reference-policy local \
  --scheduled \
  --confirm
```

sxd:
Hi @acornett21 

Well you got the issue sorry for the delay in the answer.
[...]

PR #160, opened 2024-08-12
adding call to pyxis for non EOL catalogs and updating imagestream logic

acornett21:
## Motivation
The old logic we have in place was trying to mirror `all` of the tags within a given `namespace/image`, and since there are about ~4k tags at this point, our logic would timeout and not create an ImageStream. This could cause us to be stuck on these sub-reconcilers, until a user manually added the imagestreams with the tags that they cared about.

- Fixes: #154

## Implementation
For the implementation, a graphql client for pyxis was added and a query that looks for non End Of Life catalogs. For the implementation of the query we can look for OCP versions where `eol equals null`.

### Sample GraphQL Query
```
query:
{
  find_operator_indices(
    filter: {
      and : [
        {
          organization: {
            eq: "redhat-marketplace"
          }
        },
        {
          end_of_life: {
            eq: null
          }
        }
      ]
    }
  ) {
    data{
      ocp_version
      organization
      end_of_life
      }
    error {
      status
      detail
    }
  }
}

results:
{
  "data": {
    "find_operator_indices": {
      "data": [
        {
          "ocp_version": "4.12",
          "organization": "redhat-marketplace",
          "end_of_life": null
        },
        {
          "ocp_version": "4.13",
          "organization": "redhat-marketplace",
          "end_of_life": null
        },
        {
          "ocp_version": "4.14",
          "organization": "redhat-marketplace",
          "end_of_life": null
        },
        {
          "ocp_version": "4.15",
          "organization": "redhat-marketplace",
          "end_of_life": null
        },
        {
          "ocp_version": "4.16",
          "organization": "redhat-marketplace",
          "end_of_life": null
        },
        {
          "ocp_version": "4.17",
          "organization": "redhat-marketplace",
          "end_of_life": null
        }
      ],
      "error": null
    }
  }
}
```

## Testing
This was tested in a cluster both as a catalog/bundle and also running locally in debug mode. 

```
oc get is
NAME                       IMAGE REPOSITORY                                                                       TAGS                                  UPDATED
certified-operator-index   default-route-openshift-image-registry.apps-crc.testing/oco/certified-operator-index   v4.12,v4.13,v4.14,v4.15,v4.16,v4.17   27 minutes ago
redhat-marketplace-index   default-route-openshift-image-registry.apps-crc.testing/oco/redhat-marketplace-index   v4.12,v4.13,v4.14,v4.15,v4.16,v4.17   27 minutes ago
```

As you can see we only get the tags for catalogs which are not EOL.

### DIFF
--- a/internal/pyxis/pyxis.go
+++ b/internal/pyxis/pyxis.go
@@ -1,0 +1,72 @@
+package pyxis
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/shurcooL/graphql"
+)
+
+const (
+	DefaultPyxisHost = "catalog.redhat.com/api/containers"
+)
+
+type PyxisClient struct {
+	Client    *http.Client
+	PyxisHost string
+}
+
+func (p *PyxisClient) getPyxisGraphqlURL() string {
+	return fmt.Sprintf("https://%s/graphql/", p.PyxisHost)
+}
+
+func NewPyxisClient(pyxisHost string, httpClient *http.Client) *PyxisClient {
+	return &PyxisClient{
+		Client:    httpClient,
+		PyxisHost: pyxisHost,
+	}
+}
+
+func (p *PyxisClient) FindOperatorIndices(ctx context.Context, organization string) ([]OperatorIndex, error) {
+	// our graphQL query
+	var query struct {
+		FindOperatorIndices struct {
+			OperatorIndex []struct {
+				OCPVersion   graphql.String `graphql:"ocp_version"`
+				Organization graphql.String `graphql:"organization"`
+				EndOfLife    graphql.String `graphql:"end_of_life"`
+			} `graphql:"data"`
+			Errors struct {
+				Status graphql.Int    `graphql:"status"`
+				Detail graphql.String `graphql:"detail"`
+			} `graphql:"error"`
+			Total graphql.Int
+			Page  graphql.Int
+			// filter to make sure we get exact results, end_of_life is a string, querying for `null` yields active OCP versions.
+		} `graphql:"find_operator_indices(filter:{and:[{organization:{eq:$organization}},{end_of_life:{eq:null}}]})"`
+	}
+
+	// variables to feed to our graphql filter
+	variables := map[string]interface{}{
+		"organization": graphql.String(organization),
+	}
+
+	// make our query
+	client := graphql.NewClient(p.getPyxisGraphqlURL(), p.Client)
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		return nil, fmt.Errorf("error while executing remote query for %s catalogs: %v", organization, err)
+	}
+
+	operatorIndices := make([]OperatorIndex, len(query.FindOperatorIndices.OperatorIndex))
+	for idx, operator := range query.FindOperatorIndices.OperatorIndex {
+		operatorIndices[idx] = OperatorIndex{
+			OCPVersion:   string(operator.OCPVersion),
+			Organization: string(operator.Organization),
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


internal/pyxis/types.go
```diff
@@ -0,0 +1,7 @@
+package pyxis
+
+type OperatorIndency struct {
```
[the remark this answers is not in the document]
acornett21: The struct is a singular entity. The query returns many entries.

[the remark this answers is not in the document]
komish: You're looking for `OperatorIndex` then.

internal/pyxis/pyxis.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/shurcooL/graphql"
+)
+
+const (
+	DefaultPyxisHost = "catalog.redhat.com/api/containers"
+)
+
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
```
komish, line 17: There's probably an opportunity to simplify this code here. I don't see a good use case for this interface in this code & in this context. Especially given we type assert into an http.Client and have to handle an error in that process.

internal/pyxis/pyxis.go
```diff
@@ -0,0 +1,80 @@
...
+
+type PyxisClient struct {
+	Client    HTTPClient
```
komish, line 20: If we're going to remove the interface definition, you might even embed the HTTP client here.

internal/pyxis/pyxis.go
```diff
@@ -0,0 +1,80 @@
...
+	variables := map[string]interface{}{
+		"organization": graphql.String(organization),
+	}
+
+	// make our query
+	httpClient, ok := p.Client.(*http.Client)
+	if !ok {
+		return nil, fmt.Errorf("client could not be used as http.Client")
+	}
+	client := graphql.NewClient(p.getPyxisGraphqlURL(), httpClient)
```
komish, line 64: For clarity, in another comment I reference a type assertion - and this is the offending line.

internal/pyxis/pyxis.go
```diff
@@ -0,0 +1,80 @@
...
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		return nil, fmt.Errorf("error while executing find_operator_indices query: %v", err)
```
komish, line 68: Is this error clear enough to OCO administrators?

acornett21: I've updated this lmk your thoughts.

internal/pyxis/pyxis.go
acornett21: I switched to just use an `http.Client` here directly.

internal/pyxis/pyxis.go
acornett21: I removed this and just use an `http.Client` directly.

bcrochet (changes requested):

internal/reconcilers/certified_image_stream.go
```diff
@@ -22,20 +26,29 @@ const (
 
 type CertifiedImageStreamReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log         logr.Logger
+	Scheme      *runtime.Scheme
+	PyxisClient *pyxis.PyxisClient
```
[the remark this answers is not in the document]
acornett21: Good catch, I'll remove the export. I don't want to update the signature, since this updates the interface, meaning all reconcilers would have to be updated, even though not all need this.

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #168, opened 2025-05-14
Update lint to use golangci-lint v2

caxu-rh:
- Update Makefile to get and use golangci-lint v2
- Update .golangci-lint.yaml file to match v2 config file format
- Address/ignore findings raised by staticcheck

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,71 +1,89 @@
+version: "2"
 run:
-  # Default timeout is 1m, up to give more room
+  # Default timeout is 0 (no limit), have an upper limit
   timeout: 4m
-
-linters:
-  enable:
-  - asciicheck
-  - gofumpt
-  - goimports
-  - importas
-  - ginkgolinter
-  - prealloc
-  - revive
-  - misspell
-  - stylecheck
-  - tparallel
-  - unconvert
-  - unparam
-  - unused
-  - whitespace
-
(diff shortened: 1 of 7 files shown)


acornett21:
@caxu-rh This doesn't break any of the other make cmds? ie these changes don't get overlayed/overwritten by:

```
generate
manifests
bundle
```

I guess since CI got past `generate` and `manifest` we are fine there, did you test `bundle`?

caxu-rh:
`make generate`, `make manifest`, `make bundle` all seem to work (no errors). After running these on top of the commit in this branch, the only untracked change is in `config/manager/kustomization.yaml`, which I assume is caused by me running locally, not by the lint change:

```diff
diff --git a/config/manager/kustomization.yaml b/config/manager/kustomization.yaml
index 6604f8b..4555d47 100644
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/opdev/operator-certification-operator
-  newTag: 0.0.0
+  newTag: latest
```

The `.golangci.yaml` already seems to attempt to exclude generated files, too.

acornett21:
I see that change as well, it's not caused by your change.

acornett21 (approved):

